PR #6, opened 2017-09-17
Add CharStrategy::new_borrowed.

Centril:
Adds:
+ `char::CharStrategy::new_borrowed`.

Makes it a bit more ergonomic to construct a `CharStrategy` in the more usual case (not `Owned`).


AltSysrq:
Released in version 0.3.1.

PR #8, opened 2017-11-18
Type aliases:  ValueFor<S> and W<T> + subsequence + other stuff

Centril:
This PR adds:
1. Added a type alias `ValueFor<S>` where `S: Strategy`. This is a shorter way
  to refer to: `<<S as Strategy>::Value as ValueTree>::Value`.

2. Added a type alias `type W<T> = (u32, T)` for a weighted strategy `T` in the
  context of union strategies.

3. `TestRunner` now implements `Default`.

4. Added `Config::with_cases(number_of_cases: u32) -> Config` for simpler
  construction of a `Config` that only differs by the number of test cases.

5. The default number of test cases to run is `256` (as before). This default
  which is encoded by `Config::default()` is now configurable by setting the
  env-var `PROPTEST_REQUIRED_CASES` to a value that can be parsed as a `u32`.
  This variable is read at runtime.

6. bumped dependency `rand = "0.3.18"`.

7. Added `proptest::collection::subsequence<T: Clone>(Vec<T>, Range<usize>)`
  which returns a strategy generating subsequences, of the source `Vec`,
  with a size within the given `Range`.

8. Added the trait `CollectionStrategy` which provides `.prop_sample` directly
  on collection types. The operation is the same as `subsequence` for `Vec`.

And then bumps the version to `0.3.2`.

Point 1. makes documentation quite more readable and it's a bit easier now to construct your own generic strategies.

Point 2. makes working with unions a bit more readable too, but this was less important.

I also ran `cargo clippy` and applied some of its suggestions.

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -32,7 +32,18 @@ use test_runner::*;
 // TODO: tests for CollectionStrategy impls.
 
 /// Returns a `Strategy` that generates subsequences of `source` where the
-/// where the subsequences have a size: `size.start <= seq.len() < size.end`.
+/// where the subsequences have a number of elements `len`. Given a range
+/// `size = min..max` then `min <= len < max`.
+///
+/// # Safety
+///
+/// If any of the following hold, a panic will ensue:
+///
+/// + `min > len`, since there are fewer elements than the minimum.
+/// + `max <= len`, since the max size can never happen.
+/// + `min >= max`, since the interval is empty.
+///
+/// This is equivalent to a panic ensuing iff: `!(min <= len < max)`.
 pub fn subsequence<T>(source: Vec<T>, size: Range<usize>) -> SubseqStrategy<T>
 where
     T: Clone + fmt::Debug
@@ -51,15 +62,19 @@ pub trait CollectionStrategy: Sized + fmt::Debug {
     type ValueTree: ValueTree<Value = Self>;
 
     /// Constructs a `Strategy` to generate samples of `self` given a range
-    /// (low..high) where the output sample has a number of elements: len.
-    /// The following always holds for the sample: `low <= len < high`.
+    /// `size = min..max` where the output sample has a number of
+    /// elements `len`. The following always holds for the sample:
+    /// `min <= len < max`.
     ///
     /// # Safety
     ///
-    /// A panic must occur at least when:
+    /// At least, if any of the following hold, a panic will ensue:
     ///
-    /// 1. `low > high`,
-    /// 2. `self.len() < low + 1`.
+    /// + `min > self.len()`, since there are fewer elements than the minimum.
+    /// + `max <= self.len()`, since the max size can never happen.
+    /// + `min >= max`, since the interval is empty.
+    ///
+    /// This is equivalent to a panic ensuing iff: `!(min <= len < max)`.
     fn prop_sample(self, size: Range<usize>) -> Self::Strategy;
 }
 
@@ -71,7 +86,7 @@ impl<T: Clone + fmt::Debug> CollectionStrategy for Vec<T> {
     }
 }
 
-/// A `Strategy` to generate subsequences, of a `Vec<T>`,
+/// A `Strategy` to generate subsequences, of a `seq: Vec<T>`,
 /// where the subsequences have a size: `min <= seq.len() < max`.
 #[derive(Debug, Clone)]
 pub struct SubseqStrategy<T: Clone + fmt::Debug> {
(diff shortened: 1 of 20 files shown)


AltSysrq:
Looks good to me in general. I'll likely be merging this in the near future. (Currently on vacation, so timing might be a bit weird; early next week at latest.) Thanks for the PR!

src/collection.rs
```diff
@@ -23,6 +26,143 @@ use strategy::*;
 use tuple::TupleValueTree;
 use test_runner::*;
 
+// TODO: add more CollectionStrategy impls for different types such as
+// HashSet.
+
+// TODO: tests for CollectionStrategy impls.
+
+/// Returns a `Strategy` that generates subsequences of `source` where the
```
vitiral: nit: `where the` is said twice

src/collection.rs
```diff
@@ -23,6 +26,143 @@ use strategy::*;
...
+    /// Constructs a `Strategy` to generate samples of `self` given a range
+    /// (low..high) where the output sample has a number of elements: len.
+    /// The following always holds for the sample: `low <= len < high`.
+    ///
+    /// # Safety
+    ///
+    /// A panic must occur at least when:
+    ///
+    /// 1. `low > high`,
+    /// 2. `self.len() < low + 1`.
```
vitiral: shouldn't this be `self.len() < low`? I'm not sure if a panic is nessary if `low == len`

src/collection.rs
```diff
@@ -23,6 +26,143 @@ use strategy::*;
...
+impl<T: Clone + fmt::Debug> CollectionStrategy for Vec<T> {
+    type Strategy = SubseqStrategy<T>;
+    type ValueTree = SubseqValueTree<T>;
+    fn prop_sample(self, size: Range<usize>) -> Self::Strategy {
+        SubseqStrategy::new(self, size)
+    }
+}
+
+/// A `Strategy` to generate subsequences, of a `Vec<T>`,
+/// where the subsequences have a size: `min <= seq.len() < max`.
```
vitiral: again, could this be `min <= seq.len() <= max`

src/collection.rs
```diff
@@ -23,6 +26,143 @@ use strategy::*;
...
+impl<T: Clone + fmt::Debug> SubseqStrategy<T> {
+    /// Constructs a `SubseqStrategy` which will generate subsequences of
+    /// `source` where the subsequences have a size: `[size.start, size.end)`.
+    ///
+    /// # Safety
+    ///
+    /// A panic occurs when:
+    ///
+    /// 1. `size.start >= size.end`,
+    /// 2. `source.len() < size.start + 1`.
```
vitiral: ditto

src/collection.rs
Centril: Oops; Nice catch ;)

src/collection.rs
Centril: I was unsure about this too...

but... the problem is how this works:
```rust
        // Generate the amount of elements in result:
        let rng = runner.rng();
        let amount = rand::distributions::Range::new(
            self.size.start, self.size.end).ind_sample(rng);
```

Here `amount \in [self.size.start, self.size.end)`.
Take note of the fact that `self.size.end` is never included in the interval.
So `0..10` will at most give `9`.
Please correct me if I'm wrong on this point - or how (and if) we should solve this.

src/collection.rs
Centril: See comment above.

src/collection.rs
Centril: See comment above.

src/collection.rs
vitiral: when you say `low` don't you mean `size.start`?

I would say it should panic if:
- `self.len() < size.start`
- `self.len() >= size.end`

I'm getting a little confused about the different names of these.

Centril: Oh yeah =)

A panic should happen iff: `!(size.start <= self.len() < size.end && size.start < size.end)`.

vitiral: yes, but I would break that up to be more readable :)

Also, doesn't `Range` *already* panic if `size.start > size.end`? Do we really need to specifiy it here?

Centril: To my knowledge `https://doc.rust-lang.org/nightly/std/ops/struct.Range.html` does not panic when `start > end` and it doesn't hurt to reiterate the requirement.

vitiral: ya, you're right that it doesn't hurt. I actually didn't realize that was a core type in rust.

src/collection.rs
```diff
@@ -23,6 +26,163 @@ use strategy::*;
...
+
+impl<T: Clone + fmt::Debug> ValueTree for SubseqValueTree<T> {
+    type Value = Vec<T>;
+
+    fn current(&self) -> Self::Value {
+        // We verify manually that self.len <= self.data.len(),
+        // therefore it is safe. The assertion should always hold.
+        assert!(self.len <= self.data.len());
+        unsafe {
+            slice::from_raw_parts(self.data.as_ptr(), self.len)
```
AltSysrq: Why not `self.data[..self.len].into()`?

Centril: Oh yeah, you can (and should) do that :P
You can take over from here ;)
I guess you can change this in additional commits.

Centril:
Removed the `unsafe` bit now + some more clippy. Can you take on from here?

AltSysrq:
Sorry, I thought github would add a link here or something; the PR's already merged into [`0.3.2-changes`](https://github.com/AltSysrq/proptest/tree/0.3.2-changes)

Centril:
Nicely done. Any idea when you'll merge into master? You might want to rebase two changes in `TestRunner` and `dateparser_v2` on top of `0.3.2-changes`. Also... should I close this PR? No idea how GitHub deals with it.

AltSysrq:
The only thing I have left to do for 0.3.2 is audit other places for the same kind of bug `VecStrategy` had. If all goes well, 0.3.2 will be tomorrow. I'm fairly sure that this PR will be marked as merged as soon as my branch is merged into master. If not, I'll figure something out then.

Centril:
Great =) I'll continue my work on `Arbitrary` then =)

AltSysrq:
This is all in 0.3.2, except for `CollectionStrategy` which as far as I can tell isn't able to be made useful without ATCs (see commit description in https://github.com/AltSysrq/proptest/commit/ea2008507401b7f8ad5bebff6eb163deaa0ff5d3).

PR #20, opened 2017-12-16
Change return type of new_value to compact repr + type alias for it, Add TestCaseError::{reject, fail}, reject_case, fail_case

Centril:
## Release strategy

- Since this bumps to 0.4.0 it might be prudent to add in more breaking changes while we're at it.

## What's included

- Some cleanup
- Some semantic compression
- Possible speedups by avoiding calling `case.current()` a bit
- More below..

- Instead of returning `-> Result<Self::Value, String>`, strategies are expected
  to return `-> Result<Self::Value, Rejection>` instead. `Rejection` reduces
  the amount of heap allocations, especially for `.prop_filter(..)` where
  you may now also pass in `&'static str` as well as `Rc<str>`.
  You will only experience breaks if you've written your own strategy types
  or if you've used `TestCaseError::Reject` or `TestCaseError::Fail` expliclty.

### New Additions

- Added `proptest::strategy::Rejection` which allows you to avoid heap allocation
  in some places or share allocation with string-interning.

- Added a type alias `proptest::strategy::NewTree<S>` where `S: Strategy`
  defined as: `type NewTree<S> = Result<<S as Strategy>::Value, Rejection>`.

- Added `TestCaseError::reject` and `reject_case` (short-hand).

- Added `TestCaseError::fail` and `fail_case` (short-hand).

### DIFF
--- a/src/strategy/rejection.rs
+++ b/src/strategy/rejection.rs
@@ -1,3 +1,12 @@
+//-
+// Copyright 2017 The proptest developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
 //! Provides `Rejection`, a compact representation of why something, such
 //! as a generated value, was rejected.
 
--- a/src/strategy/rejection.rs
+++ b/src/strategy/rejection.rs
@@ -1,0 +1,185 @@
+//! Provides `Rejection`, a compact representation of why something, such
+//! as a generated value, was rejected.
+
+use std::borrow::Borrow;
+use std::cmp::Ordering;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::ops::Deref;
+use std::rc::Rc;
+
+use self::Rejection::*;
+
+/// The reason for why something, such as a generated value, was rejected.
+///
+/// # Representation
+/// 
+/// This type is currently (representationally) equivalent to:
+///
+/// ```rust
+/// pub enum Rejection {
+///     Borrowed(&'static str),
+///     BoxOwned(Box<str>),
+///     RcOwned(Rc<str>),
+/// }
+///
+/// # fn main() {
+/// # use std::mem::size_of;
+/// assert_eq!(size_of::<Rejection>(), 3 * size_of::<usize>());
+/// # }
+/// ```
+///
+/// This is a compact and essentially copy-on-write representation while also
+/// allowing you not to allocate if the rejection message is not dynamic in
+/// nature. This is most efficient and should be used when possible. If you
+/// are allocating and then cloning the allocated `Rejection`, then you should
+/// prefer the `Rejection::RcOwned` variant as it will only bump the reference
+/// count and not actually do any new heap allocation.
+pub enum Rejection {
+    /// The borrowed representation - this is the most efficient if you for
+    /// example have a string in the data section of your program.
+    Borrowed(&'static str),
+    /// A reference counted shared string, use this if you have dynamically
+    /// generated content that is used more than once.
+    RcOwned(Rc<str>),
+    /// An owned boxed string, use this if you have a dynamicly generated
+    /// string that is only used once. If used more than once, instead use
+    /// `RcOwned`.
+    BoxOwned(Box<str>),
+}
+
+/// Constructs and returns a [`Rejection`] based on the input type.
+/// See the `From<T> for Rejection` implementations for details.
+///
+/// # Example
+///
+/// ```rust
+/// fn main() {
+///     use proptest::strategy::reject;
+///     let reason = format!("The value {:?} was too much!", 100);
+///     let reject = reject(reason);
+///     println!("{:?}", reject);
+/// }
+/// ```
+///
+/// [`Rejection`]: type.Rejection.html
+pub fn reject<S: Into<Rejection>>(string: S) -> Rejection {
+    string.into()
+}
+
+impl Rejection {
+    /// Converts this `Rejection` into either an owned one backed by `Rc`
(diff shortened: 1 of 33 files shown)


src/strategy/rejection.rs
```diff
@@ -0,0 +1,196 @@
...
+/// # }
+/// ```
+///
+/// This is a compact and essentially copy-on-write representation while also
+/// allowing you not to allocate if the rejection message is not dynamic in
+/// nature. This is most efficient and should be used when possible. If you
+/// are allocating and then cloning the allocated `Rejection`, then you should
+/// prefer the `Rejection::RcOwned` variant as it will only bump the reference
+/// count and not actually do any new heap allocation.
+pub enum Rejection {
```
AltSysrq: This is eerily similar in functionality to [Supercow](https://docs.rs/supercow/0.1.0/supercow/). I'm not sure whether we'd want to bring that into proptest, just noting.

Centril: The `Rc` part? What would be the downside? Carrying around `String` is also 3 words but we don't use the `capacity` word for anything atm. Removing `Rc` still means that `Rejection` is 3 words, but you loose the ability to cheaply clone. I think `Rc` is primarily useful in `Filter` where cloning of the string can happen frequently.

src/strategy/rejection.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// The borrowed representation - this is the most efficient if you for
+    /// example have a string in the data section of your program.
+    Borrowed(&'static str),
+    /// A reference counted shared string, use this if you have dynamically
+    /// generated content that is used more than once.
+    RcOwned(Rc<str>),
```
AltSysrq: If this case weren't here, `Rejection` could be replaced by `Cow<'static, str>`. I don't think the cost of allocating upon every rejection in what's currently the `Rc` case is really worth having a unique smart pointer type.

src/test_runner.rs
```diff
@@ -157,17 +157,47 @@ pub enum TestCaseError {
...
+    }
+}
+
+/// Short-hand for `Err(TestCaseError::reject(..))`.
+pub fn reject_case<R: Into<Rejection>>(reason: R) -> TestCaseResult {
+    Err(TestCaseError::reject(reason))
+}
+
+/// Short-hand for `Err(TestCaseError::fail(..))`.
+pub fn fail_case<R: Into<Rejection>>(reason: R) -> TestCaseResult {
```
AltSysrq: These two are largely redundant with `prop_assert!` and `prop_assume!`, since both will virtually always be nested inside a condition.

src/strategy/rejection.rs
AltSysrq: > The Rc part?

The whole thing really. Supercow is a reference + box + arc/rc/etc smart pointer, `Rejection` is reference + box + rc smart pointer.

AltSysrq:
I do appreciate the cleanup here, and fixing the thing where one had to write `.prop_filter("foo".to_owned(), ...)` is a nice improvement. Aside from what I mentioned above, :+1: for the PR. Thanks as always!

src/strategy/rejection.rs
Centril: I first considered using [`Bow`](https://docs.rs/bow/1.0.3/bow/enum.Bow.html) since `assert_eq!(32, size_of::<Cow<'static, str>>());` while `assert_eq!(24, size_of::<Bow<'static, str>>());`. But `Filter` does a lot of `.clone()` on `whence` so I included `Rc` as well. With `&'static str` we can avoid allocating at all. What's the downside besides compile times when the interface doesn't become harder to use?

src/test_runner.rs
Centril: That's fair =) Can we make it non-`pub` for now? It adds some semantic compression inside `proptest` itself at places.

Centril:
Can we perhaps do the following?
- Hide `fail_case` / `reject_case` for the time being as internal detail.
- Keep `Rejection` but hide the internal representation behind a `struct` so the variants are non-matchable outside of the crate and construction is only possible via `From` and friends. Then we have the freedom to change the representation later without breaking changes.

AltSysrq:
> `fail_case` / `reject_case`

I went through the changes here, and these two don't really seem to save that much space.

`reject_case` has two usages, one inside `prop_assume!` and the `gives_up_after_too_many_rejections` test that uses it unconditionally.

`fail_case` has 8 use sites, but 6 of them could be replaced with a use of the `prop_assert!` macro. The other two are `prop_assert!` itself and `panic_guard`.

> Keep `Rejection` but hide the internal representation behind a `struct`

That sounds better, but I'm still not convinced it's worth the internal complexity. The main advantage for `Rejection` (compared to, say, putting generics on `prop_filter` so it could also take an `&str`) is that you don't need to allocate every time a test case is rejected, right? The thing is, the default budget for local rejections is 65536, and the rejection messages are fairly small. Even if `String::clone()` took a full microsecond (I just benchmarked 46ns on my system for a 49-char string), that's only 66ms of CPU time burned before the test runner bails anyway.

Centril:
So I can remove `reject_case` and `fail_case` if you want =)

But that is 66ms per test right? With 100 tests, a not too unreasonable amount, that is 6.6 seconds. Tho I assume that a (very) low percentage of those 100 tests will actually hit so many rejections so perhaps 0.6 seconds (guessing wildly, maybe you have a more informed guesstimate here).

Hmm, if not this... Can we use `Cow<'static, str>`? And at least whatever we do hide the representation behind a type that is not `String` so we can change in the future if we wish?

AltSysrq:
> But that is 66ms per test right?

If you hit this number, you're either running hundreds of thousands if not millions of test cases and the rejection time is insignificant, or else you're rejecting the vast majority of generated inputs and should probably rethink the strategy.

That said, I pretty much agree entirely with the compromise you proposed in the last comment. Even ignoring the performance concerns, having the possibility of supporting something richer than a string in `TestCaseError::Reject` would be nice.

Centril:
Fixed things as discussed.

For some inexplicable reason, GitHub seems to think there are still merge conflicts...

PR #22, opened 2017-12-17
Add failing case persistence.

AltSysrq:
A solution to https://github.com/AltSysrq/proptest/issues/21. [How it works](
https://github.com/AltSysrq/proptest/blob/378e430ad926e596f68f7df34d06a478b87541e5/README.md#failure-persistence)

I've taken the approach of making this "just work" by default with options to configure/disable it if desired, since I can't think of any cases where this would actually introduce breakage.

@Centril Any opinions? There's probably some stuff like filenames that might be worth bike-sheding too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,17 @@ thread 'main' panicked at 'Test failed: byte index 4 is not a char boundary; it 
 '
 ```
 
-The first thing we should do is copy the failing case to a traditional unit
-test since it has exposed a bug.
+The first thing we should do is add the new [failure
+persistence](#failure-persistence) file to our source control, for example
+with git:
+
+```text
+$ git add src/proptest-failures.txt
+```
+
+The next thing we should do is copy the failing case to a traditional unit
+test since it has exposed a bug not similar to what we've tested in the
+past.
 
 ```rust
 #[test]
@@ -270,7 +279,12 @@ Between those two, though, we see something different: it tried to shrink
 the `0000-06-20` and `0000-09-20` test cases _passed_.
 
 In the end, we get the date `0000-10-01`, which apparently gets parsed as
-`0000-00-01`. Again, let's add this as its own unit test:
+`0000-00-01`. Again, this failing case was added to the failure persistence
+file, and we should add this as its own unit test:
+
+```text
+$ git add src/proptest-failures.txt
+```
 
 ```rust
 #[test]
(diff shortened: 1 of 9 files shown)


Centril:
Some thoughts:
- Does this work well with `.prop_flat_map` and `.prop_perturb`?

README.md
```diff
@@ -120,8 +120,17 @@ thread 'main' panicked at 'Test failed: byte index 4 is not a char boundary; it
...
 ```
 
-The first thing we should do is copy the failing case to a traditional unit
-test since it has exposed a bug.
+The first thing we should do is add the new [failure
+persistence](#failure-persistence) file to our source control, for example
+with git:
+
+```text
+$ git add src/proptest-failures.txt
```
Centril: - Perhaps the default folder should be `src/regressions/..` ?

AltSysrq:
> `prop_perturb`

As well as could be hoped; it's fully deterministic.

> `prop_flat_map`

Good catch, right now it makes a new RNG with `rand::weak_rng()`; I'll need to change that.

src/test_runner.rs
```diff
@@ -107,10 +113,20 @@ pub struct Config {
     /// The default is 1_000_000, which can be overridden by setting the
     /// `PROPTEST_MAX_FLAT_MAP_REGENS` environment variable.
     pub max_flat_map_regens: u32,
+    /// Indicates how to determine the file to use for persisting failed test
+    /// results.
+    ///
+    /// See the docs of [`FailurePersistence`](enum.FailurePersistence.html)
+    /// for more information.
+    ///
+    /// The default is `FailurePersistence::WithMain("proptest-failures.txt")`.
```
Centril: Could each `#[test]` be given its own file by default? Perhaps not... It is very frustrating that we have `module_path!()` but not `function!()` to determine the name of the function... Can the `proptest!` macro perhaps use the `ident` for the test-function name and `stringify!(..)` it and then append that?

AltSysrq: The main part of the reasoning for having all the tests use one file is that I found that I often have things like

```rust
proptest! {
  fn test_a(x in foo_strategy()) { ... }
  fn test_b(x in foo_strategy()) { ... }
}
```

and if something breaks one test, there's a reasonable chance it could also break the other test.

Maybe a happy medium would be one failures file per source file? E.g., from `src/foo/bar.rs` you'd get something like `src/foo/bar-proptest-failures.txt`. That would also solve the question of "where to put this one global file by default?".

Centril: I think that's a very reasonable compromise and you can use `module_path!()` directly then =)

Tho I think it might be good not to put the file directly in the same folder as the module and instead have a special fixture folder, like so:

```
src/foo/bar.rs => src/regressions/foo/bar-proptest-failures.txt
```
or:
```
src/foo/bar.rs => regressions/foo/bar-proptest-failures.txt
```

AltSysrq:
Implemented the file location changes, including defaulting to the separate directory tree. Persisting the files alongside the sources with a different extension is also an option, mainly to act as a fallback if proptest can't figure out your source tree.

`prop_flat_map()` is deterministic now, too.

Centril:
Really cool work =)

Yesterday a (possibly bad) thought occurred to me: What if we **in addition** allow the user to specify a function of type `fn(Seeds, OtherInterestingData) -> io::Result<()>` and let them run arbitrary code that might for example ship the seeds of to some log-database somewhere and then let them optionally provide a (sorta) inverse function `fn() -> (Seeds, OtherInterestingData)` for getting back the seed if they so with... All of this can then be abstracted behind a trait `FailurePersister`.

Is this too much perhaps? Arguably they can still do this with `inotify` and custom scripts over the files changed.

AltSysrq:
> Is this too much perhaps?

I'd say yes for now. Not in the sense that it's something I definitely don't want to add, but I'd rather have a very concrete use-case before adding something like that. Otherwise there's a substantial risk of creating an abstraction in the API that's awkward or impossible to use for anything other than what proptest does internally.

Centril:
Right; Then I have nothing more to add - nice work and let's merge? =)

PR #23, opened 2017-12-17
Support u128 and i128 behind unstable Cargo.toml flag for nightly users.

Centril:
## In this PR

- Added support for `i128` and `u128`. Since this is an unstable feature in
  Rust, this is hidden behind the feature `unstable` which you have to
  explicitly opt into in your `Cargo.toml` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased
+
+### New Additions
+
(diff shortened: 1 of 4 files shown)


AltSysrq:
:+1: Nothing I can complain about here.

Centril:
@hellow554  should be fixed once we move to the next rand version; cc https://github.com/AltSysrq/proptest/issues/107

PR #25, opened 2018-01-02
fixes #24 location of failure peristence file for workspace member crates

### LINKED ISSUE #24: Failure persistence file location for workspace member crates
I've just started using proptest. It looks great so far. The project I'm using it is organized as a Cargo workspace with several member crates. When using proptest in a member crate of a workspace the location of failure persistence file is not what I expected.

The project structure is like this:
```
my_workspace/
|-> Cargo.toml    #defines the workspace
|-> my_member_crate/
    |-> Cargo.toml
    |-> src/lib.rs
    |-> src/protocol.rs   #code with tests module
```
in this case the failure persistence file is located at:
`my_workspace/my_member_crate/my_member_crate/protocol.protest-regressions`

what I expected is:
`my_workspace/my_member_crate/protest-regressions/protocol.txt`

I have played around with the source code of test_runner. I found that the source_path is relative to the workspace root, e.g. my_member_crate/src/lib.rs. But the working directory when running the tests is the crate root, so my_member_crate in this case.

haraldmaida:
I'm not sure if this is the perfect solution but it works for me now.

### DIFF
--- a/src/test_runner.rs
+++ b/src/test_runner.rs
@@ -19,7 +19,7 @@ use std::fmt;
 use std::fs;
 use std::io::{self, BufRead, Write};
 use std::panic::{self, AssertUnwindSafe};
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 use std::sync::{Arc, RwLock};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::SeqCst;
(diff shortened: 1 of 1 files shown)


AltSysrq:
Thanks for the PR, though I opted to fix the root issue by determining the absolute path to the source earlier on, which also fixes a related problem in the other configuration options.

The tests probably passed for you but failed on travis because the test that checked how the persistence file was located were not run on Windows (:man_facepalming:). That's fixed now as well.

PR #27, opened 2018-01-14
PR #20, mergable + some changes

AltSysrq:
@Centril Not sure what happened with the merge in your PR, but I managed to branch off of your last commit before that and got something git is happy with. There's also a couple extra changes here to `Rejection`, mainly renaming it to `Reason` and making it not be a smart pointer so that new methods/fields can be added to it in the future.


Centril:
Looks good =)

PR #32, opened 2018-01-21
Fixes compiler errors in unstable tests

MasonRemaley:
Hey, thanks for the great library!

I was messing around with it a bit and found that when I ran `cargo test --features unstable` locally, I got a bunch of compiler errors due to `feature(i128_type)` being enabled in the library but not `feature(i128)`, and `feature(i128_support)` not being enabled in the `rand` dependency.

This PR fixes that, once fixed all the unstable tests do still pass. I assume that they used to compile, but things changed and weren't caught cause travis doesn't enable the unstable feature in the nightly tests. I was thinking it'd probably be a good idea to change that so that this wouldn't happen again in the future but haven't yet figured out how to do that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ travis-ci = { repository = "AltSysrq/proptest" }
 default = []
 
 # Enables unstable features of Rust.
-unstable = []
+unstable = ["rand/i128_support"]
 
 [dependencies]
 bitflags = "1.0.1"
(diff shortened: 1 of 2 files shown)


AltSysrq:
Thanks for the fix!

> I think travis doesn't enable the unstable feature in the nightly tests. I was thinking it'd probably be a good idea to change that

Yeah, I don't see a clear way to do that either.

AltSysrq:
Released in 0.4.2.

MasonRemaley:
No prob, if I find a way to set features in Travis I'll make another PR.

PR #33, opened 2018-01-22
Travis test unstable

MasonRemaley:
I think I might've figured out how to test with unstable enabled on travis, I couldn't figure out how to test it with travis on my fork so this PR is a work in progress for now until I see if Travis does what I expect, will reply to this again once I get it working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ rust:
 - nightly
 
 matrix:
+  include:
+    - rust: nightly
+      env: FEATURES=unstable
   allow_failures:
     - rust: nightly
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
   allow_failures:
     - rust: nightly
 
+# Install travis-cargo which sets `--feature unstable` for nightly builds automatically
 before_script:
   - |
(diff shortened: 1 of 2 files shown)


MasonRemaley:
Ah turns out that's not really what that environment variable does, I'll mess with this some more later.

MasonRemaley:
Ah nice thanks for the links! I'll try setting it up the way you have it in `proptest-arbitrary`.

MasonRemaley:
Awesome, this seems to be working  now!

Btw is there a reason to do `cargo build; cargo test` instead of just `cargo test`? I set this up to do the former since that's what you have in `proptest-arbitrary` but I'm not sure what the difference is.

MasonRemaley:
Neat, it actually turns out that `travis-cargo` automatically sets a  feature named unstable when appropriate. Works now!

Centril:
Perhaps we should not allow failures on nightly and require that a commit be made fixing those changes?
It would be nice if failures were more noticeable.

AltSysrq:
@MasonRemaley Thanks for the contribution!

@Centril Agreed, I'll make that change after merging.

MasonRemaley:
No problem!

PR #38, opened 2018-01-26
don't pin dependencies in Cargo.toml

### LINKED ISSUE #37: exporting Rng may have caused other problems (sorry)
after updating proptest I'm getting this error:

```
error[E0277]: the trait bound `R: rand::Rng` is not satisfied
  --> src/test/implemented.rs:43:9
   |
43 |         regex_generate::Generator::parse(r"\.([a-zA-Z0-9_]{1,10})", rng.clone()).unwrap();
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `rand::Rng` is not implemented for `R`
   |
   = help: consider adding a `where R: rand::Rng` bound
   = note: required by `regex_generate::Generator`

```

I'm trying to trace down what is going wrong, but it seems like `extern crate rand; rand::Rng` is being treated differently than `proptest::Rng`, even though they are the _same version_. I'm not totally sure what's going on

vitiral:
Originally I thought this would fix #37 (it doesn't, that was because of regex_generate)

I think it is still considered best practices to not pin dependencies on a library. Let me know if you have other thoughts.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,12 +24,12 @@ default = []
 unstable = ["rand/i128_support"]
 
 [dependencies]
-bitflags = "1.0.1"
-bit-set = "0.4.0"
-quick-error = "1.2.1"
-rand = "0.4.2"
-regex-syntax = "0.4.2"
-lazy_static = "1.0.0"
+bitflags = ">=1.0.1"
+bit-set = ">=0.4.0"
+quick-error = ">=1.2.1"
+rand = ">=0.4.2"
+regex-syntax = ">=0.4.2"
+lazy_static = ">=1.0.0"
 
 [dev-dependencies]
-regex = "0.2.5"
+regex = ">=0.2.5"


Centril:
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

Cargo interprets what is specified right now as:
```toml
[dependencies]
bitflags = "^1.0.1"
bit-set = "^0.4.0"
quick-error = "^1.2.1"
rand = "^0.4.2"
regex-syntax = "^0.4.2"
lazy_static = "^1.0.0"

[dev-dependencies]
regex = "^0.2.5"
```

which is another way of saying:

```toml
[dependencies]
bitflags = ">= 1.0.1 <2.0.0"
bit-set = ">= 0.4.0 <0.5.0"
quick-error = ">= 1.2.1 <2.0.0"
rand = ">= 0.4.2 <0.5.0"
regex-syntax = ">= 0.4.2 <0.5.0"
lazy_static = ">= 1.0.0 <2.0.0"

[dev-dependencies]
regex = ">= 0.2.5 <0.3.0"
```

Which I believe is the right thing semver wise.

AltSysrq:
I agree with @Centril, and this is also what other crates do: [quickcheck](https://github.com/BurntSushi/quickcheck/blob/master/Cargo.toml#L24), [rand](https://github.com/rust-lang-nursery/rand/blob/master/Cargo.toml#L28-L51), [serde](https://github.com/serde-rs/serde/blob/master/serde_derive/Cargo.toml#L25).

One thing proptest does do differently is specify the full version rather than the minimally-theoretically-usable version. I'm not sure how much that matters in general, but it wouldn't have changed the original problem here either.

Centril:
@AltSysrq On pinning the patch version; Do we have a good rationale for that? Shouldn't that only be done when you know that the patch version is needed to fix some problem we have?

killercup:
> I think it is still considered best practices to not pin dependencies on a library. Let me know if you have other thoughts.

I have other thoughts! :)

As @Centril wrote, Cargo by default treats a "1.0" as "everything semver-compatible with 1.0" (so >= 1.0 and <2.0). Specifying the lowest supported version and trusting the crate author and cargo to fetch compatible updates is the best practice IMHO.

You should only ever use custom ranges for crates you _publicly_ expose and that you know are compatible. (Bumping from ">= 0.1, <0.3" to ">= 0.1, <0.4" can be done in a minor/patch version update.)

killercup:
> the greater range is what I am suggesting we use.

Sorry, I'm not sure I follow. What greater range? I was suggesting to start with "0.1" and only make it range when (a) you publicly expose items from the dependency in your crate, and (b) the version are compatible.

I would _never_ use an unbounded version range like ">=1.0.1" in a library.

vitiral:
This current PR doesn't reflect what I want to change so I'm closing

I would like to open a new PR that uses the full range of what this library can support. For instance `>=1.0.1 < 3.0` for _[insert dependency here]_.

I would like to get tests running that validate the version range is _actually_ supported first though 😄

PR #39, opened 2018-02-09
make (S)BoxedStrategy into newtypes

Centril:
CHANGELOG

- `BoxedStrategy` and `SBoxedStrategy` are now newtypes instead of being type
  aliases. You will only experience breaking changes if you've directly
  used `.boxed()` and not `(S)BoxedStrategy<T>` but rather
  `Box<Strategy<Value = Box<ValueTree<Value = T>>>>`. The probability of
  breakage is very small, but still possible. The benefit of this change
  is that calling `.boxed()` or `.sboxed()` twice only boxes once. This can
  happen in situations where you have functions `Strategy -> BoxedStrategy` or
  with code generation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+## Unreleased
+
+### Potential Breaking Changes
+
+- `BoxedStrategy` and `SBoxedStrategy` are now newtypes instead of being type
+  aliases. You will only experience breaking changes if you've directly
+  used `.boxed()` and not `(S)BoxedStrategy<T>` but rather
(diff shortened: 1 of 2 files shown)


AltSysrq:
@Centril Just to double-check, does https://github.com/AltSysrq/proptest/commit/fce5b1333935c4624ac7e2b43a1164a93faa16a6 look sane?

Centril:
@AltSysrq LGTM =)

AltSysrq:
Published in 0.5.0

PR #40, opened 2018-02-11
Reduce heap allocation in .prop_recursive(..) + make it more ergonomic

Centril:
## Other changes

+ Bugfix in `#[cfg(target_os = "windows")] fn osstring_invalid_string()`. Since this was unreleased it's not included in the CHANGELOG.

+ Some housekeeping (reorganising) + documentation fixes.

- There is a small chance of breakage if you've relied on the constraints put
  on type inference by the closure in `leaf.prop_recursive(..)` having a fixed
  output type. The output type is now any strategy that generates the same type
  as `leaf`. This change is intended to make working with recursive types a bit
  easier as you no longer have to use `.boxed()` inside the closure you pass to
  `.prop_recursive(..)`.

### Minor changes

- Reduced the heap allocation inside `Recursive` somewhat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ### Potential Breaking Changes
 
+- There is a small chance of breakage if you've relied on the constraints put
+  on type inference by the closure in `leaf.prop_recursive(..)` having a fixed
+  output type. The output type is now any strategy that generates the same type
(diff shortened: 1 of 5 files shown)


AltSysrq:
Thanks! I should be able to get all this published this weekend.

AltSysrq:
Published in 0.5.0

PR #43, opened 2018-03-06
ArcStrategy (and arc_boxed) + Recursive improvements + housekeeping

Centril:
## Housekeeping

- Some general house keeping in `test_runner.rs` and `flatten.rs`.

### New Additions

- Added `ArcStrategy<T>` and `.arc_boxed()` for boxed strategies affording
  cheap shallow cloning by reference counting (internally using an `Arc`).

- There is a small change of breakage if you've relied on `Recursive` using
  a `BoxedStrategy` as `Recursive` now internally uses `ArcStrategy<T>`
  instead as well as expecting a `Fn(ArcStrategy<T>) -> R` instead of
  `Fn(Arc<BoxedStrategy<T>>) -> R`.

### Minor changes

- Reduced indirections and heap allocations inside `Recursive` somewhat.

### Bug Fixes

- Removed `impl Arbitrary for LocalKeyState` since `LocalKeyState` no longer
  exists in the nightly compiler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+## Unreleased
+
+### New Additions
+
+- Added `ArcStrategy<T>` and `.arc_boxed()` for boxed strategies affording
+  cheap shallow cloning by reference counting (internally using an `Arc`).
+
+### Potential Breaking Changes
+
+- There is a small change of breakage if you've relied on `Recursive` using
+  a `BoxedStrategy` as `Recursive` now internally uses `ArcStrategy<T>`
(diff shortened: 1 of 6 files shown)


AltSysrq:
👍 Looks good in general.

Thought: Is there any reason not to make `BoxedStrategy` just use `Arc` internally? Since it's now hidden behind a newtype, you can't do anything particular to `Box` anymore. And since nobody allocates extremely large number of strategies, the overhead of `Arc` vs `Box` should be pretty minimal.

AltSysrq:
If you'd like to do that yourself, be my guest; otherwise I'll just take care of it on Sat before publishing the new version.

PR #45, opened 2018-03-09
Separate out the ValueTree of Filter<S, F> into FilterValueTree + reduce cloning

Centril:

- `Filter<S, F>` and `statics::Filter<S, F>` no longer implement `ValueTree`.
  They now delegate this responsibility to the new types `FilterValueTree` and
  `statics::FilterValueTree`. The risk of breakage is however small.

### Minor changes

- Reduced the amount of cloning that `Filter<S, F>` (and the `static::` cousin)
  does by introducing `FilterValueTree`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   `Fn(BoxedStrategy<T>) -> R`. In addition, the type of recursive strategies
   has changed from `Recursive<BoxedStrategy<T>, F>` to just `Recursive<T, F>`.
 
+- `Filter<S, F>` and `statics::Filter<S, F>` no longer implement `ValueTree`.
+  They now delegate this responsibility to the new types `FilterValueTree` and
+  `statics::FilterValueTree`. The risk of breakage is however small.
+
 ### Minor changes
 
 - Reduced indirections and heap allocations inside `Recursive<T, F>` somewhat.
@@ -16,6 +20,9 @@
   using `Box`. While this has marginal overhead, it also reduces the overhead
   in `Recursive<T, F>`. The upside to this change is also that you can very
   cheaply clone strategies.
+
+- Reduced the amount of cloning that `Filter<S, F>` (and the `static::` cousin)
+  does by introducing `FilterValueTree`s.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


AltSysrq:
I was originally planning on merging this, but I don't really like adding another inconsistency to the type names (currently `Shuffle` and `Perturb` are the only 1:1 adaptors that have a different value tree type), and I'd like to keep the number of "breaking changes for a very small performance improvement" entries in the change log to a minimum (even for changes like this that probably don't actually break any real code) for the sake of perception.

The main benefit of this PR can also be had by just using

```rust
    whence: "unused".into(),
```

to ensure no allocation happens. It's definitely a bit of a smell to have an unused field there, but I think it's tolerable.

I have the change I mentioned above committed locally, but I'd like to hear any thoughts you have before I go forward down that path.

AltSysrq:
Aside, I'm really hoping "impl trait in trait" comes soon, which should make all of this moot

Centril:
> I'd like to keep the number of "breaking changes for a very small performance improvement" entries in the change log to a minimum (even for changes like this that probably don't actually break any real code) for the sake of perception.

Understandable =)

> to ensure no allocation happens. It's definitely a bit of a smell to have an unused field there, but I think it's tolerable.

Yeah; I think it is.

> Aside, I'm really hoping "impl trait in trait" comes soon, which should make all of this moot

I assume this is referring to:

```rust
impl<S: Strategy, F: ..> Strategy for Filter<S, F> {
    existential type Value: ValueTree<Value = ValueFor<S>>;
    ...
}
```

This is not implemented in nightly, and so I highly doubt it will be in epoch 2018. It's not in the current planning of WG-epoch at least.

What we could do is to hide `FilterValueTree` from the documentation and possibly not even export it to users; This will make transitioning to `existential type` (or whatever the final syntax is) not require another breaking change on account of this in the future.

On a side note I'd like to keep the type aliases `FilterMinSize<S, Mapper>` since they make the internal implementation a bit more readable =)

AltSysrq:
> What we could do is to hide `FilterValueTree` from the documentation and possibly not even export it to users

Hm, what would this look like in the docs? I like the idea in concept, I just don't want to send people on a wild goose chase looking for a hidden type referenced in the docs or compiler error messages.

AltSysrq:
It was necessary to pubish a new version anyway to update the nightly features. 0.6.0 includes the alternate approach we discussed above, so I'm closing this PR since I think the immediate issue is addressed. Feel free to reopen if I missed something.

PR #46, opened 2018-03-11
Refactor test_runner.rs as a folder

### LINKED ISSUE #9: Long term plan: Investigate strategies for non-owned types
Currently, it is impossible to have a `S: Strategy` where `ValueOf<S>` is non-owned.

Examples:
1. We have a base strategy `T` where `ValueOf<T>` is owned, and then we use that to make a strategy where `ValueOf<S> = &'a ValueOf<T>` - This will not type check.
2. Derived types, for example: `Cow<'a, X>`.

Generally, I think that, any type that has a lifetime can't be generated.

Being able to generate types with lifetimes in them is however useful when you have lifetimes somewhere deep in the hierarchy of a type. To be able to do this soundly, the `ValueTree` must perhaps have a notion of `Owned` (which defaults to `Value`) in addition to `Value` (which may be borrowed). Then `TestRunner` will return the `Owned` type instead of `Value`, but the function under test receives `Value`.
I tried to implement this and got `TestRunner` to compile, but no strategies would compile.
Solving this may require generic associated types (GAT). I'll experiment with it once we have GAT in nightly.

There's also `*const T` and `*mut T`. While it is possible to give a sound strategy for these by [`Box::leak`](https://github.com/rust-lang/rust/pull/45881), this may cause problems since tests for large data types may leak huge amounts of memory. This could perhaps be mitigated by letting the `ValueTree` store the previous value of `.current()`, and whenever `.current()` happens, the old value is deallocated. The last value needs to be persisted in the `TestRunner` somehow if the entire `ValueTree` did not fail so that it can be cleared.

I'd like your thought's on the desirability of these features.

Centril:
This PR is a pure refactoring and should have zero semantic or public changes.

--------------------

I was originally planning on changing the signature of `run` from:

```rust
pub fn run<S, F>(&mut self, strategy: &S, test: F)
    -> Result<(), TestError<ValueFor<S>>>
where
    S : Strategy,
    F : Fn (&ValueFor<S>) -> TestCaseResult
```

to:

```rust
pub fn run<S, F>(&mut self, strategy: &S, test: F)
    -> Result<(), TestError<S::Value>> // <-- Note that the ValueTree is returned
where
    S : Strategy,
    F : Fn(ValueFor<S>) -> TestCaseResult // <-- Note the absence of & here.
```

but right now it didn't seem worth it to make that change. However - I think we should do that change once GATs land (and try it out in nightly) since it should allow us to address #9 as well as giving users an owned value in their tests without having to clone test case values (and not all types implement clone..).


AltSysrq:
Thanks for this, it's increasingly felt necessary but I never got around to it.

FTR, your PR description contains the text `fix #9`, which github interprets to mean that merging this PR will close that issue; I'm going to take the liberty to edit that to something else to avoid close/reopen noise.

Centril:
Oh thanks, I forgot about that =)

PR #48, opened 2018-04-25
Add no_std + alloc support, refactoring persistence

ZackPierce:
Pursuant to https://github.com/AltSysrq/proptest/issues/47

Three related kinds of changes were necessary:

* The failure seed persistence system was made into a generic trait, `FailurePersistence`, of which `FileFailurePersistence` (the legacy enum of file-lookup-related-variants) is one available option.
* Innumerable uses of standard library code were either replaced with core-based alternatives, often conditionally.
* New features were added to manage conditional compilation and dependency features: "std", "nightly", and "alloc". Currently the only sensible combinations are either "std" (the default) or "nightly" and "alloc" but not "std."

I'm not a fan of the verbosity necessary to manage all the conditional uses, but I didn't see any non-magical alternative.

A few nonessential changes may have snuck in due to auto-rustfmt'ing. If any are problematic, I'd be happy to roll them back.

CC: @AltSysrq


src/test_runner/failure_persistence/map.rs
```diff
@@ -0,0 +1,108 @@
...
+#[cfg(feature = "std")]
+use std::vec::Vec;
+
+#[cfg(all(feature = "alloc", not(feature="std")))]
+use alloc::{BTreeMap, BTreeSet};
+#[cfg(feature = "std")]
+use std::collections::{BTreeMap, BTreeSet};
+
+/// Failure persistence option that loads and saves seeds in memory
+/// on the heap.
```
AltSysrq: Could you expand in the docs on how this would be useful?

ZackPierce: Sure thing! Updated in an additional commit.

ZackPierce:
Some additional fixes had to be made to account for changes in `nightly` that removed access to some float-related methods (see: https://github.com/rust-lang/rust/pull/49896 ) .  All checks are green now.

ZackPierce:
@AltSysrq Ping for prospective approval/merge/version-bump/glory.

AltSysrq:
Sorry for the delay. Latest changes look good, so I'll see about getting this out now.

ZackPierce:
Awesome, thank you!

PR #54, opened 2018-06-05
Update regex-syntax dependency

ZackPierce:
I realized we'd have to update our regex-syntax dependency if we were ever planning on pulling in the upcoming no_std + alloc regex capabilities, so here we have a switch from the old, pinned regex-syntax to a more recent version.

The most obvious difference is that in the updated regex-syntax, the crate provides a higher-order representation that is further digested than the raw AST.   It accomplishes useful things like resolving the effects of case insensitivity by modifying the character/byte classes presented.

The largest implication, AFAICT, from using this more digested data source is that we lose some ability to provide custom behavior for special-case character classes (e.g. `AnyCharNoNL`), without doing some significantly deeper inspection of our own. As a result, the prior characteristic of manually overweighting certain character ranges in our generated output has been dropped.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -14,9 +14,9 @@ use std::fmt;
 use std::u32;
 
 #[cfg(all(feature = "alloc", not(feature="std")))]
-use alloc::borrow::{Cow, ToOwned};
+use alloc::borrow::Cow;
 #[cfg(feature = "std")]
-use std::borrow::{Cow, ToOwned};
+use std::borrow::Cow;
 
 #[cfg(all(feature = "alloc", not(feature="std")))]
 use alloc::boxed::Box;
@@ -24,9 +24,9 @@ use alloc::boxed::Box;
 use std::boxed::Box;
 
 #[cfg(all(feature = "alloc", not(feature="std")))]
-use alloc::string::String;
+use alloc::string::{String, ToString};
 #[cfg(feature = "std")]
-use std::string::String;
+use std::string::{String, ToString};
 
 #[cfg(all(feature = "alloc", not(feature="std")))]
 use alloc::vec::Vec;
@@ -38,8 +38,6 @@ use regex_syntax as rs;
 use bool;
 use char;
 use collection;
-use bits;
-use num;
 use strategy::*;
 use test_runner::*;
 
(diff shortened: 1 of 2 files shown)


src/string.rs
```diff
@@ -125,112 +123,89 @@ pub fn string_regex_parsed(expr: &rs::Expr)
...
+                        |r: &rs::hir::ClassUnicodeRange | (r.start(), r.end())).collect();
+                    Ok(char::ranges(Cow::Owned(ranges))
+                        .prop_map(to_bytes).sboxed())
+                },
+                Class::Bytes(ref class) => {
+                    let subs = (class).iter().map(
+                        |r: &rs::hir::ClassBytesRange | if 255u8 == r.end() {
+                            (r.start()..).sboxed()
+                        } else {
+                            (r.start()..r.end()).sboxed()
```
AltSysrq: Should this be `r.end() + 1`? (Though I notice the old code also just uses `start..end`)

src/string.rs
```diff
@@ -125,112 +123,89 @@ pub fn string_regex_parsed(expr: &rs::Expr)
...
+    match *hir.kind() {
+        HirKind::Empty => Ok(Just(vec![]).sboxed()),
+        HirKind::Literal(ref l) => {
+            match *l {
+                rs::hir::Literal::Unicode(ref ch) => {
+                    Ok(Just(ch.to_string()
+                        .into_bytes()).sboxed())
+                },
+                rs::hir::Literal::Byte(ref byte) => {
+                    Ok(Just(::core::iter::once(*byte).collect()).sboxed())
```
AltSysrq: Is there a reason not to use `vec![*byte]`?

AltSysrq:
https://travis-ci.org/AltSysrq/proptest/jobs/388072089

Looks like this'll need to wait for the 0.8.0 train in which support for Rust 1.21 is dropped.

ZackPierce:
Roger that, I'll be thinking through strategies for keeping the manual weighting capabilities comparably applied.

I think we can probably get this thing working in 1.21 by being more explicit with references.

Centril:
Should probably close this now? It's already included in version 0.8.0.

PR #55, opened 2018-06-05
General refactor

Centril:
# Other changes

- general cleanup...

# User facing changes

### New Additions

- `i128` and `u128` are now supported without any feature flags and on stable.

- More implementations of `Arbitrary` are supported for `alloc` + `no_std` users.

- `size_range` now accepts inclusive ranges of form `low..=high` and `..=high`.
  Thus, you can construct a `vec` strategy as: `vec(elt_strategy, low..=high)`
  and `vec(elt_strategy, ..=high)`. This also applies to other functions
  accepting `Into<SizeRange>`.

- `..= high` is now a valid strategy. Please note that `..= 1` will include
  numbers lower than `0`. If you want the range `0..=1`, then you have to write
  `..=1u<size>` where `<size>` is one of `u8`, `u16`, `u32`, `u64`, `usize`.

- `low..=high` is also a valid strategy except for `f32` and `f64` which will
  soon be supported in newer editions (1.27+) of Rust.

- `Arbitrary` is implemented for `RangeInclusive<Idx>`, `RangeToInclusive`,
  and `DecodeUtf16` on stable.

### Minor changes

- The Bernoulli distribution is now used for `proptest::bool::weighted`.

### Breaking changes

- A minimum version of 1.26 of Rust is now required.

- `regex-syntax` version 0.6 is now used.

- `rand` version 0.5 is now used.

- As a consequence, the stored seed is now `[u8; 16]` instead of `[u32; 4]`.
  To minimize breakage for failure persistence, proptest will until version 0.9
  accept both the new and old seed format in persisted files.
  When you run proptest on version 0.8, it will automatically convert your
  persisted files to the new format.

- The RNG used by proptest has been changed to a PRNG `TestRng` which proptest
  exposes. This is currently a simple new-type wrapper around `XorShiftRng`.
  In the future, this will give us more freedom to make changes without breakage.

- The feature flag `i128_support` has been removed. The features it added are
  now always supported.

- The associated type `Value` of `Strategy` has been renamed to `Tree`.
  A new associated type `Value` has been added to `Strategy` which always refers
  to the same type as `<S::Tree as ValueTree>::Value` for some strategy `S`.
  This change allows you to write `-> impl Strategy<Value = T>` for functions
  returning a `Strategy` generating `T`s. This is more ergonomic to use than
  `-> impl Strategy<Value = impl ValueTree<Value = T>>`.

- As a consequence change, the associated type `ValueTree` has been removed from
  `Arbitrary`.

- The methods `run` and `run_one` on `TestRunner` now takes a function-under-test
  that accepts the generated type by value instead of by reference instead.
  This means that you don't need to write `ref value in my_strategy` and can
  write `value in my_strategy` instead even if `typeof(value)` doesn't implement
  `Copy`. This is also a step in the direction of allowing strategies to generate
  references when generic associated types (GATs) land.
  However, `ref value in my_strategy` will still be accepted, so not a lot of
  breakage should come of this if you've used `proptest! { .. }`.

- `prop_compose!` no longer applies `.boxed()` to the strategy produced.
  Therefore, `-> BoxedStrategy<T>` is no longer the correct type.
  The new return type is `-> impl Strategy<Value = T>`.
  If you want the old behaviour, you can use `.boxed()` yourself.

- `Arbitrary` for `SizeRange` changed its associated type to use `RangeInclusive`.
  Same applies for `CString`.

### Nightly-only breakage

- As `std::io::{Chars, CharsError}` have been deprecated on nightly,
  their `Arbitrary` implementations have been removed.


Centril:
Still not done:
+ I'd like to rename `new_value` to `new_tree` as a consequence change of renaming the associated type to `Tree`.

Todo before next version:

+ I am working on `#[derive(Arbitrary)]` again after troubles with recursive types; I'd like to ship that as well before landing this in a new version.

Centril:
Also note that this branch conflicts with #54 since it also does that change but with some different logic. I think I can reuse some parts of the other PR tho (like tests).

examples/dateparser_v1.rs
```diff
@@ -32,18 +32,22 @@ fn parse_date(s: &str) -> Option<(u32, u32, u32)> {
...
     }
 }
 
 fn main() {
-    assert_eq!(None, parse_date("2017-06-1"));
-    assert_eq!(None, parse_date("2017-06-170"));
-    assert_eq!(None, parse_date("2017006-17"));
-    assert_eq!(None, parse_date("2017-06017"));
-    assert_eq!(Some((2017, 6, 17)), parse_date("2017-06-17"));
+    #[cfg(feature = "std")]
```
AltSysrq: Since the examples are intended for people to read, I'd prefer sticking with "just don't build the examples on no-std"

src/array.rs
```diff
@@ -99,8 +99,9 @@ macro_rules! small_array {
             }
         }
 
-        impl<S : Strategy> Strategy for [S;$n] {
-            type Value = ArrayValueTree<[S::Value;$n]>;
+        impl<S : Strategy> Strategy for [S; $n] {
+            type Tree = ArrayValueTree<[S::Tree; $n]>;
+            type Value = [ValueFor<S>; $n];
```
AltSysrq: Isn't `ValueFor` essentially obsolete now?

src/bits.rs
```diff
@@ -16,17 +16,20 @@
...
 use core::ops::Range;
+use std_facade::fmt;
 
 use bit_set::BitSet;
 use rand::{self, Rng};
 
 use strategy::*;
 use test_runner::*;
 
+// FIXME(2016-06-04): Consistently use Into<SizeRange>
```
AltSysrq: 2018

src/num.rs
```diff
@@ -706,13 +762,13 @@ macro_rules! float_bin_search {
                     };
                     let new_mid = self.lo + interval/2.0;
 
-                    let new_mid = if new_mid == self.curr || 0.0 == interval {
+                    let new_mid = if nearly_eq(new_mid, self.curr) || 0.0 == interval {
```
AltSysrq: I'm fairly confident the strict equality is correct. Was this just changed because of the clippy lint?

Centril: Yes; the tests do pass however :)

So... could we document why the strict equality is correct here?

src/array.rs
Centril: Kinda; but keeping it around causes less breakage? And if we make future changes, then the alias helps reduce it further.

But I'll remove it if you want.

Centril:
Fixed the outstanding issues (I hope).
Did the deprecation and renaming here as well (sorry about that..);
I'll make smaller PRs henceforth :)

Centril:
So, the regex `.` produces partly the following regex HIR:

```rust
ranges: [
    ClassUnicodeRange {
        start: "0x0",
        end: "0x9"
    },
    ClassUnicodeRange {
        start: "0xB",
        end: "\u{10ffff}"
    }
]
```

This corresponds exactly to `AnyCharNoNL`. We can trivially check for that exact range sequence and then apply the special logic to that. If we agree this is a good idea, I'll fix this in a separate PR.

src/string.rs
```diff
@@ -85,192 +88,205 @@ impl Strategy for str {
...
+    })
 }
 
-fn flip_ascii_case(flip: bool, ch: u8) -> u8 {
-    if flip && ch >= b'a' && ch <= b'z' {
-        ch - b'a' + b'A'
-    } else if flip && ch >= b'A' && ch <= b'Z' {
-        ch - b'A' + b'a'
+fn to_bytes(khar: char) -> Vec<u8> {
+    // Logic from: https://doc.rust-lang.org/nightly/src/core/char/methods.rs.html#438-468
```
AltSysrq: This whole function could be reduced to

```rust
let mut buf = [0u8;4];
khar.encode_utf8(&mut buf).as_bytes().to_owned()
```

src/test_runner/failure_persistence/file.rs
```diff
@@ -144,159 +233,118 @@ where
...
-                            .display(),
-                        err
-                    );
-                }
-                vec![]
-            }
+fn convert_to_new_format(old_format: [u32; 4]) -> [u8; 16] {
+    use byteorder::{NativeEndian, ByteOrder};
+    let mut new_format = [0; 16];
+    NativeEndian::write_u32_into(&old_format[..], &mut new_format);
```
AltSysrq: Does `rand` actually convert the seed into `[u32;4]` using native byte order? If so, it'd probably be better to keep storing the `u32`s since otherwise they're not portable.

src/test_runner/failure_persistence/file.rs
```diff
@@ -144,159 +233,118 @@ where
...
-                let mut ret = Vec::new();
+    // Split by whitespace and ignore empty lines:
+    let parts = line.trim().split(char::is_whitespace).collect::<Vec<_>>();
+    let len = parts.len();
+    if len == 0 { return None; }
+
+    // Decide which parser to use depending on format:
+    let first_ok = parts[0] == "xs";
+    let parser = match len {
+        5 if first_ok => {
```
AltSysrq: The point of the leading `xs` was to be able to handle things like this, so it would be better to change the type.

src/test_runner/failure_persistence/file.rs
```diff
@@ -144,159 +233,118 @@ where
...
+            eprintln!("proptest: using old format parsing [u32; 4]. \
+                       This format will be removed in version 0.9");
```
AltSysrq: What is the user's recourse here? I'd rather support the old format indefinitely.

src/string.rs
Centril: Fixed.

src/test_runner/failure_persistence/file.rs
Centril: Don't think rand does that; It seems to convert back into `[u32; 4]` using little endian; https://docs.rs/rand/0.5.0/src/rand/prng/xorshift.rs.html#89-130 (`rand` uses little endian everywhere now apparently...)

`le::read_u32_into` is defined here: https://github.com/rust-lang-nursery/rand/blob/591519b5a6d5b8ae14f7bb536cb3439449ad8092/rand_core/src/le.rs#L18-L39

src/test_runner/failure_persistence/file.rs
Centril: My understanding is that when proptest is run, then the old format will be read in, and then saved to the same file using the new format, so the migration should be automatic?

src/test_runner/failure_persistence/file.rs
Centril: Not sure I follow what you mean by *"change the type"*...?

src/test_runner/failure_persistence/file.rs
AltSysrq: > the old format will be read in, and then saved to the same file using the new format

The regression file is append-only.

> Are you saying that the old format should be supported indefinitely parallel with the new format

Yes, that.

src/test_runner/failure_persistence/file.rs
AltSysrq: `xs` is the type of the line. So `xs` would mean "four u32s" and we create a new line type for the new format.

AltSysrq: (The name is for `XorShift` because I didn't anticipate the type of the seed ever changing.)

src/test_runner/failure_persistence/file.rs
AltSysrq: I suppose if we support the old format indefinitely, there's not really a benefit for the new one.

src/test_runner/failure_persistence/file.rs
Centril: Oh; now I get it! I didn't realize `xs` stood for `XorShift` at all ;)

Sounds like a good idea to change the prefix 👍

src/test_runner/failure_persistence/file.rs
Centril: > The regression file is append-only.

Ah I see; I read the code incorrectly before.

> I suppose if we support the old format indefinitely, there's not really a benefit for the new one.

I suppose not;

My hope was to drop the `byte-order` dependency in the not too distant future by not supporting the old format beyond 0.9 so that we don't accumulate technical debt.

Would it be feasible to offer a separate program that performs the migration automatically perhaps?
i.e: If you point it to a `failure-persistence.txt` file, it will convert the file to the new format.

AltSysrq: > My hope was to drop the `byte-order` dependency in the not too distant future

I wouldn't be opposed to just having the two conversions in private functions in this crate, though the `byteorder` crate also doesn't seem like a particularly heavy dependency.

Centril: So... I think these two definitions of `convert_to_new_format` should be equivalent; `convert_to_new_format` is the definition using the exact same logic as in `byteorder`.

https://play.rust-lang.org/?gist=0347f1b9fbf316aae1a85e48c6d79e4d&version=stable&mode=debug

Centril:
Per discussion I dropped the new `[u8; 16]` format in persistence files.
I also kept the `byteorder` dependency.

The last commit does some small consistency fixes in documentation I forgot about.

PR #56, opened 2018-06-05
Weigh AnyCharNoNL in old regex-syntax properly in regex-syntax 0.6 logic

Centril:
This PR depends on #55.

Previously in #54: 
> the prior characteristic of manually overweighting certain character ranges in our generated output has been dropped

This is fixed by this PR. The prior characteristics should be retained.


Centril:
Rewrote the logic a bit to check in both directions for the `AnyCharNoNL` range set; `regex-syntax` doesn't generate it now, but it makes things more robust.

PR #62, opened 2018-06-25
Permit specifying property tests as closures

udoprog:
This change permits easily sharing things from the environment.

Consider a simple test framework like this:

```rust
let expensive_thing = setup_expensive_thing();

let runner = TestRunner::new();
runner.test("my test", proptest! {
    |(x in 0u32..10000)| {
        expensive_thing.do_something(x)?;
    }
});
```

### DIFF
--- a/src/sugar.rs
+++ b/src/sugar.rs
@@ -83,22 +83,7 @@ macro_rules! proptest {
                 let mut config = $config.clone_with_source_file(file!());
                 config.test_name = Some(
                     concat!(module_path!(), "::", stringify!($test_name)));
-                let mut runner = $crate::test_runner::TestRunner::new(config);
-                let names = proptest_helper!(@_WRAPSTR ($($parm),*));
-                match runner.run(
-                    &$crate::strategy::Strategy::prop_map(
-                        proptest_helper!(@_WRAP ($($strategy)*)),
-                        |values| $crate::sugar::NamedArguments(names, values)),
-                    |$crate::sugar::NamedArguments(
-                        _, proptest_helper!(@_WRAPPAT ($($parm),*)))|
-                    {
-                        $body;
-                        Ok(())
-                    })
-                {
-                    Ok(_) => (),
-                    Err(e) => panic!("{}\n{}", e, runner),
-                }
+                proptest_helper!(@_BODY config ($($parm in $strategy),+) $body);
             }
         )*
     };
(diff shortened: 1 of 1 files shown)


src/sugar.rs
```diff
@@ -111,6 +96,20 @@ macro_rules! proptest {
...
     } };
+
+    (|($($parm:pat in $strategy:expr),+)| $body:block) => {
+        || {
+            let config = $crate::test_runner::Config::default();
+            proptest_helper!(@_BODY config ($($parm in $strategy),+) $body);
+        }
+    };
+
+    (move |($($parm:pat in $strategy:expr),+)| $body:block) => {
```
Centril: Given this...

src/sugar.rs
```diff
@@ -1071,3 +1089,17 @@ mod ownership_tests {
...
+mod closure_tests {
+    #[test]
+    fn test_simple() {
+        let x = 42;
+
+        let _ = proptest! {
+            |(y in 0..100)| {
+                assert!(x != y);
+            }
+        };
```
Centril: ... add a test that ensures that it keeps working if we make changes :)

src/sugar.rs
```diff
@@ -674,6 +673,25 @@ macro_rules! proptest_helper {
     (@_WRAPSTR ($a:pat, $($rest:pat),*)) => {
         (stringify!($a), proptest_helper!(@_WRAPSTR ($($rest),*)))
     };
+    // build a property testing block that when executed, executes the full property test.
+    (@_BODY $config:ident ($($parm:pat in $strategy:expr),+) $body:block) => {{
```
AltSysrq: The code that sets the source file should probably be here so that it also applies to the new case.

src/sugar.rs
```diff
@@ -111,6 +96,20 @@ macro_rules! proptest {
         $($(#[$meta])*
           fn $test_name($($parm in $strategy),+) $body)*
     } };
+
+    (|($($parm:pat in $strategy:expr),+)| $body:block) => {
+        || {
+            let config = $crate::test_runner::Config::default();
```
AltSysrq: We should check that the config doesn't have `fork()` enabled since that feature requires proptest having control over the top-level test function.

udoprog: Thanks!

Ended up checking for both `fork` and `timeout` features, even though one depends on another. Possibly worth to introduce a `no_fork` flag in config which takes precedence to make this clearer.

src/sugar.rs
udoprog: Done. I kept the `.clone()` in the calling macro to match how it behaved with `clone_with_source_file()`.

src/sugar.rs
udoprog: Thanks, fixed. This one slipped my mind!

Centril:
PS: before you release this; I'm working on a PR related to it :)

AltSysrq:
This is released in 0.8.1.

PR #64, opened 2018-06-27
Change regression file line parse

### LINKED ISSUE #63: Possible regression in 0.8.0
I just started using proptest in a brand new project, and after getting my first failure, subsquent runs were failing to parse the regressions file and new failures were not being added to the existing file:

During the test run, proptest output this during the test run:
```
failures:
 
---- tests::round_trip stdout ----
        proptest: /home/project/proptest-regressions/types/time/period.txt:1: unknown case type `` (corrupt file or newer proptest version?)
proptest: /home/project/proptest-regressions/types/time/period.txt:2: unknown case type `` (corrupt file or newer proptest version?)
proptest: /home/project/proptest-regressions/types/time/period.txt:3: unknown case type `` (corrupt file or newer proptest version?)
proptest: /home/project/proptest-regressions/types/time/period.txt:4: unknown case type `` (corrupt file or newer proptest version?)
proptest: /home/project/proptest-regressions/types/time/period.txt:5: unknown case type `` (corrupt file or newer proptest version?)
proptest: /home/project/proptest-regressions/types/time/period.txt:6: unknown case type `` (corrupt file or newer proptest version?)
```
The first six lines are the commented lines. It is likely due to the code [here](https://github.com/AltSysrq/proptest/blob/ba8d5f476dfde5fbd3d1442d8a66155799b33235/src/test_runner/failure_persistence/file.rs#L243), as split will return one part with the entire string if there is nothing to split on.

While I had never used proptest 0.7.2, I downgraded to that version, and using the same file, was able to create a new failure that added cases to the regressions file.

neoeinstein:
Split will always leave at least one part, even if there is no delimeter. This corrects the logic to use the string's length to check for emptiness before trying to split it.

(PR created with the in-browser GitHub file editor)

Fixes #63

### DIFF
--- a/src/test_runner/failure_persistence/file.rs
+++ b/src/test_runner/failure_persistence/file.rs
@@ -237,10 +237,10 @@ fn parse_seed_line(mut line: String, path: &Path, lineno: usize)
         line.truncate(comment_start);
     }
 
-    // Split by whitespace and ignore empty lines:
-    let parts = line.trim().split(char::is_whitespace).collect::<Vec<_>>();
-    let len = parts.len();
-    if len > 0 {
+    if line.len() > 0 {
+        // Split by whitespace and ignore empty lines:
+        let parts = line.trim().split(char::is_whitespace).collect::<Vec<_>>();
+        let len = parts.len();
         // "xs" stands for "XorShift".
         if parts[0] == "xs" && len == 5 {
             // Parse using the chosen one:


Centril:
Mea culpa! Thank you for spotting this. ❤️

AltSysrq:
Thanks for the fix!

PR #65, opened 2018-06-27
Allow `pattern:type` in `proptest!(..)` and `prop_compose!(..)` + documentation fixes

Centril:
This PR aims to allow you to write:

```rust
proptest! {
    #[test]
    fn addition_is_commutative(a: u8, b in 0u8..) {
        ...
    }
}
```

and such things... The diff is a bit longer than I would have liked, but Rust won't let me define a matcher: `$p:pat : $s:ty` since `':' \in FOLLOW(pat)`.

It also fixes some documentation inaccuracies in `sugar.rs`.

### DIFF
--- a/src/sugar.rs
+++ b/src/sugar.rs
@@ -18,8 +18,7 @@ use std_facade::fmt;
 ///
 /// Each function will be wrapped in a function which sets up a `TestRunner`,
 /// and then invokes the function body with inputs generated according to the
-/// strategies. Note that the inputs are borrowed from the test runner, so if
-/// they are not `Copy`, you will need to use `ref` with each parameter name.
+/// strategies.
 ///
 /// Example:
 ///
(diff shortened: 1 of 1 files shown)


AltSysrq:
> Rust won't let me define a matcher: `$p:pat : $s:ty` since `':' \in FOLLOW(pat)`.

Yeah, that's why the existing syntax uses `in` in the first place (the fact that it meshes nicely with this syntax is a rather happy accident). It's so weird, since that sequence occurs so frequently in usual Rust syntax.

Anyway, I'm definitely in favour of this. Since I did promise to get some stuff out for 0.8.1 tonight and the appveyor build takes so long, this'll probably slip to a separate 0.8.2 release this weekend.

src/sugar.rs
```diff
@@ -112,6 +162,22 @@ macro_rules! proptest {
             proptest_helper!(@_BODY config ($($parm in $strategy),+) $body);
         }
     };
+
+    (|($($arg:tt)+)| $body:block) => {
```
AltSysrq: You'll want to pull in similar changes to https://github.com/AltSysrq/proptest/commit/c85573f4ed42ea1df350ff33eb1d8737b233e3a8 here

Centril:
> Yeah, that's why the existing syntax uses in in the first place (the fact that it meshes nicely with this syntax is a rather happy accident). It's so weird, since that sequence occurs so frequently in usual Rust syntax.

Yup. I'm hoping we can do some really interesting stuff with proc macros to write `#[proptest]` directly on functions and such things eventually :)

I fixed the conflicts; should hopefully be ready to go (or not, if you want me to improve anything?).

AltSysrq:
LGTM. I'm going to put this on the 0.8.2 train though so I can get tonight's bug fix out.

Centril:
Neat!

AltSysrq:
Released in 0.8.2.

PR #76, opened 2018-07-23
Fix typo in tutorial

caipre:
The text refers to a range between 0 and 100, but the example was written as `0..10i32`.


AltSysrq:
Good catch, thanks!

PR #77, opened 2018-07-25
Relax requirements for generic conversion to TestCaseError

udoprog:
This permits using alternative error types which doesn't implement `std::error::Error`, like [`failure::Error`](https://docs.rs/failure/0.1.1/failure/struct.Error.html) directly in proptest blocks.

This is accomplished by relaxing the requirements on the blanket conversion mechanism to only require `fmt::Display`. Unfortunately this would cause `TestCaseError` to conflict with the blanket implementation, so it had to be reworked.

### DIFF
--- a/src/test_runner/errors.rs
+++ b/src/test_runner/errors.rs
@@ -54,11 +54,25 @@ impl TestCaseError {
     pub fn fail(reason: impl Into<Reason>) -> Self {
         TestCaseError::Fail(reason.into())
     }
+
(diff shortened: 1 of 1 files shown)


AltSysrq:
The breaking change is pretty unfortunate, though this otherwise looks like a good thing to have. For now, I'm going to keep this open under consideration for the next time we have a breaking release.

matthew-russo:
This got deleted when i switched from a `master` to `main` branch. I've cut a ticket to track this at #507

Issue #36, opened 2018-01-25
Idea: Making a book and splashy website about proptest


Centril:
Some inspiration:
+ serde.rs <-- for books
+ tokio.rs
+ rocket.rs <-- this just looks amazing.

Centril:
For now I've registered `proptest.rs` with `istanco.com` and setup DNS as follows:

| Name         | Type      | TTL | Target |
| ------------|---------| -----|-------|
| proptest.rs | A           | 300 | 192.30.252.154 |
| proptest.rs | A           | 300 | 192.30.252.153 |
| www          | CNAME | 300 | proptest.github.io |
| proptest.rs | NS         | 300 | ns1.istanco.com |
| proptest.rs | NS         | 300 | ns2.istanco.com |

AltSysrq:
I would have preferred if you hadn't spent your money predicated on me making a decision.

> a book

What would you suggest would go here? The longer tutorials would be a good fit I suppose, but is there a good way to do that without loosing easy testability of the examples?

> splashy website

I'm personally not a big fan of these. I'm also utterly incompetent at making websites. If you want to go ahead and put something together though, go ahead.

Centril:
> I would have preferred if you hadn't spent your money predicated on me making a decision.

Hehe =P I got carried away ;)

I think we should initially start with just having a gh-pages site that just contains `cargo doc` stuff and then we can make something better later.

> What would you suggest would go here?

As you say.. the longer tutorials, which would include notes on how you use custom derive, more case studies (longer examples), etc. I think `serde.rs` is what we should aspire too with the book, it seems pretty nice.

It seems that mdBook supports running tests: https://rust-lang-nursery.github.io/mdBook/cli/test.html

> I'm personally not a big fan of these. I'm also utterly incompetent at making websites. If you want to go ahead and put something together though, go ahead.

I'm pretty decent at design (I think..), but I haven't done any web programming for some time. But that would give me a great reason to write some code using rocket.rs ;)
But let's make a great book first which I'd be happy to work on ^,-

AltSysrq:
> mdBook supports running tests

Good to know.

> let's make a great book first

Sounds good to me. I unfortunately have some other stuff going on at the moment so I probably won't be able to do any serious work, but if you get something on a PR I can definitely review+merge. Hopefully things will be back to normal for me by next week though.

matthew-russo:
Closing this one as the book is published here: https://altsysrq.github.io/proptest-book/intro.html
[...]

PR #79, opened 2018-07-31
The proptest_derive crate & #[derive(Arbitrary)]

Centril:
This PR adds the crate `proptest_derive` to this repo.

The advantage of not having this in a separate repo is that I expect both crates will be easier to maintain and test if they are in a single place. Another crate that does this is [serde](https://github.com/serde-rs/serde) which sports its `serde_derive` crate [here](https://github.com/serde-rs/serde/tree/master/serde_derive).

The PR is pretty massive in size (sorry about that) but it is 100% consisting of additions and no code is touched outside of the `proptest-derive` directory. I recommend checking out my branch and starting from the `tests` directory to see what is supported and what is explicitly rejected.

I'm sure there are places where the code can be refactored and improved, but I'd like to avoid making too many changes to this PR. We can of course make design changes and code improvements later.

Some things TODO after merging this PR (to a separate branch, not `master`):

1. Setup a workspace. Ideally, all the code for the `proptest` crate should go in its own directory in the root and `proptest-derive` should also be in the root.

2. Make sure `proptest-derive` is tested with Travis CI and AppVeyor CI.  For now, I have not made any changes to `.travis.yml` but there are plenty of tests (+50% of the code). These should all be run with a nightly compiler (because `compiletest_rs` wants it for `pretty` tests and because `attr_literals` is unstable). Until we've ensured that tests are run, it would be inadvisable to make breaking changes to `proptest` itself.

3. I have to give you perms on crates.io so you can publish updates to `proptest-derive` at the same time as `proptest`.

4. Release a 0.1 of `proptest-derive`.

Here are some TODOs into the future:

1. We should figure out what to do with self-recursive and mutually-recursive types. These are not supported at all and will blow the stack if attempted.

2. Write extensive documentation (ideally in the form of an mdbook (see #36)). I haven't had the time to do that yet but I'll start on it ASAP.

3. Improve error messages with correct spans and such things.

4. Eventually use the custom error message API in https://internals.rust-lang.org/t/custom-error-diagnostics-with-procedural-macros-on-almost-stable-rust/8113 to segregate errors by warnings and so on...


AltSysrq:
Sorry for my delayed response here. I'll see if I can work through this this weekend.

> Some things TODO after merging this PR

All sound like a good plan to me.

Centril:
No problem ❤️  Take your time! :)

proptest-derive/src/ast.rs
```diff
@@ -0,0 +1,712 @@
...
+///         ))
+///     ))
+/// )
+/// ```
+///
+/// However, we can do better by being linear for the `10 - 1` first
+/// strategies and then switch to nesting like so:
+///
+/// ```ignore
+/// (1, 2, 3, 4, 5, 6, 7, 8, 9,
```
AltSysrq: This nesting isn't going to produce a uniform distribution. This is why `prop_oneof!` falls back to boxing everything so it can use a normal `Union` if it's over the limit.

AltSysrq: Oh, I see you're doing something with weights below.

proptest-derive/tests/params.rs
```diff
@@ -0,0 +1,123 @@
...
+struct InnerNoParams {
+    string: String,
+    #[proptest(no_params)]
+    has: TopHasParams,
+}
+
+#[derive(Debug, Arbitrary)]
+#[proptest(params(u64))]
+struct TPIS {
+    #[proptest(strategy = "\"a+\"")]
```
AltSysrq: How hard would it be to have a `#[proptest(regex = "string")]` for `String` and `Vec<[u8]>` fields? It would make syntax like this nicer

AltSysrq:
> These [tests] should all be run with a nightly compiler

Does the plugin itself work with stable?

I've made a pass through the code and it initially looks pretty good, though there's a lot I don't understand since this stuff is extremely new to me. I'll try to work through it in more detail this week.

proptest-derive/src/ast.rs
Centril: Yeah, the idea is that:
```rust
prop_oneof![ w1 => e1, w2 => e2, w3 => e3 ]
```
can be rewritten in terms of:
```rust
prop_oneof![ w1 => e1, (w2 + w3) => prop_oneof![ w2 => e2, w3 => e3 ] ]
```
and it commutes, so you can also say:
```rust
prop_oneof![ (w1 + w2) => prop_oneof![ w1 => e1, w2 => e2 ], w3 => e3 ]
```

This is then applied for larger groupings (the max number that `TupleUnion` supports - 1).

In practice this would differ from the distribution you get when using `Union` + boxing because you are sampling the PRNG more times; but it seems to be that the cost of boxing every part of the union outweighs this slight discrepancy so I'd like to avoid the boxing.

proptest-derive/tests/params.rs
Centril: Not very hard I think; you could do this by adding another attribute `regex` to the parser and then desugaring that to the same internals as `strategy` uses in the attribute parser.

I can do this in a follow up PR if you like :)

(I agree `#[proptest(strategy = "\"a+\"")]` looks suboptimal)

Centril:
> Does the plugin itself work with stable?

It should; I think I have tested it on stable a few times, but I don't remember exactly when the last time was, so it is possible I let some mistake slip through.

To make sure everything beside `attr_literal` (which is getting stabilized soon, https://github.com/rust-lang/rust/pull/53044) tests work on stable, I think it would be a good idea to follow this PR up with one that gates the nightly-only tests behind an `unstable` cargo feature. We can then use the same testing mechanism we use for stable/unstable in proptest proper.

Centril:
Did you get a chance to review any more of this?

AltSysrq:
I haven't. Sorry about that and how unresponsive I've been in general lately. I should have some time soon.

Centril:
Hey :)

I'd love to make improvements to the `proptest_derive` crate (for example upgrade the `syn` version and other changes) but I'd also like to avoid making too many changes to allow you to review the crate more easily.

Any chance you could review this? If you think it's unlikely that you'll get to it in the foreseeable future, I can publish this out of tree, but that will of course make dealing with breaking chances more difficult.

AltSysrq:
I'm going to set myself a hard commitment this time so I stop letting this slide: I will review and likely merge this on Saturday.

proptest-derive/src/error.rs
```diff
@@ -0,0 +1,512 @@
...
+// Messages
+//==============================================================================
+
+macro_rules! mk_err_msg {
+    ($code: ident, $msg: expr) => {
+        concat!(
+            "[proptest_derive, ", stringify!($code), "]",
+            " during #[derive(Arbitrary)]:\n",
+            $msg,
+            " Please see: https://PATH/TO/foo#", stringify!($code),
```
AltSysrq: We'll need to update this once the location of the docs are selected. (Just making a note so this is easier to see later.)

AltSysrq:
Merging this into the `proptest-derive` branch. I'll get the workspace and CI set up after that, and if you don't already have docs written somewhere, I'll start on that too.

proptest-derive/src/error.rs
Centril: Nicely spotted; I might have forgotten about this otherwise. =P

AltSysrq:
Is there a reason some of the `.rs` files have the execute bit set?

```
...
 create mode 100644 proptest-derive/tests/compile-fail/E0026-strategy-malformed.rs
 create mode 100755 proptest-derive/tests/compile-fail/E0027-filter-malformed-expr.rs
 create mode 100755 proptest-derive/tests/compile-fail/E0027-filter-malformed.rs
 create mode 100755 proptest-derive/tests/compile-fail/E0028-skipped-variant-has-filter.rs
 create mode 100644 proptest-derive/tests/compile-fail/E0028-skipped-variant-has-param.rs
 create mode 100644 proptest-derive/tests/compile-fail/E0028-skipped-variant-has-strat.rs
...
```

Centril:
@AltSysrq probably just my editor being silly (using Windows can be painful...). Try running the tests without them and change to 644 or something if the tests all pass... :)

> if you don't already have docs written somewhere, I'll start on that too.

No sorry, never got around to writing more docs but I'd like to help out; we can take some inspiration from serde.rs; an mdbook might be fitting here...

AltSysrq:
Never mind, figured it out

Centril:
@AltSysrq Hmm... this happened to me sometimes when I was using the wrong path when building locally; so my guess would be one of: a) that the `target/` directory ends up in a wrong place, b) compile_test looks in the wrong directory, c) the stuff needed are not in the right directory (configured in `https://github.com/AltSysrq/proptest/blob/proptest-derive/proptest-derive/tests/compiletest.rs`). It's like a) or b)...

Does this build locally with the branch tested or do you also get errors with that?

I'm about to sleep now, so I'll check it out in the morning if you didn't solve it before :)
...lol :D

Centril:
@AltSysrq so... what was the problem? =P

AltSysrq:
https://github.com/AltSysrq/proptest/commit/353b52f8b70e953f6814eb39406605789b5b3162

`target` is shared by the whole workspace so the directory got moved when I set that up.

Centril:
Aaah; that makes a lot of sense. :)

PR #81, opened 2018-08-10
Make StringParam constructible

dtrebbien:
Attempting to construct a StringParam from a regex pattern would lead to a compiler error with the message "constructor is not visible here due to private fields". The StringParam's sole field needs to be made public to allow external code to construct a StringParam to be passed to any_with::&lt;String&gt;.


Centril:
There is a `From<&'static str> for StringParam`, https://github.com/AltSysrq/proptest/blob/6725df39a77dc90c44978676e968af42cdb221e8/src/string.rs#L40-L42, that you can use.

dtrebbien:
Thank you. That works.

PR #84, opened 2018-08-22
[WIP] State machine support

AltSysrq:
Something I've been working on here and there for a while, originally suggested in https://github.com/AltSysrq/proptest/issues/28.

As noted in the first commit, this absolutely needs more tests and is probably still a bit deficient on documentation. I wanted to get this up now though to solicit opinions on the model/ergonomics now that the implementation is basically functional.

I'd recommend starting by reading the (still incomplete) [tutorial](https://github.com/AltSysrq/proptest/blob/9b542078ddb7ae07bbda78fdc555b5d927c75f33/src/state_machine.rs#L90) and then the syntax docs above it.

@Centril I'd appreciate your thoughts when you have time.

### DIFF
--- a/examples/state_machine_v1.rs
+++ b/examples/state_machine_v1.rs
@@ -1,0 +1,201 @@
+//-
+// Copyright 2018 Jason Lingle
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+// The code here is adapted from
+// https://hypothesis.works/articles/rule-based-stateful-testing/
+
+#[macro_use]
+extern crate proptest;
+
+use std::cmp;
+
+/// A hand-rolled implementation of a binary heap, like
+/// https://doc.rust-lang.org/stable/std/collections/struct.BinaryHeap.html,
+/// except slow and buggy.
+#[derive(Clone, Debug)]
+pub struct MyHeap<T> {
+    data: Vec<T>,
+}
+
+impl<T : cmp::Ord> MyHeap<T> {
+    pub fn new() -> Self {
+        MyHeap { data: vec![] }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.data.is_empty()
+    }
+
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &T> {
+        self.data.iter()
+    }
+
+    pub fn push(&mut self, value: T) {
+        self.data.push(value);
+        let mut index = self.data.len() - 1;
+        while index > 0 {
+            let parent = (index - 1) / 2;
+            if self.data[parent] < self.data[index] {
+                self.data.swap(index, parent);
+                index = parent;
+            } else {
+                break;
+            }
+        }
+    }
+
+    pub fn pop(&mut self) -> Option<T> {
+        // THIS IS WRONG
+        if self.is_empty() {
+            None
+        } else {
+            Some(self.data.swap_remove(0))
+        }
+    }
+}
+
+mod test {
+    use std::fmt;
+
+    use super::*;
+    use proptest::prelude::*;
+    use proptest::state_machine::StateMachine;
+
+    // These would normally be in `#[cfg(test)]` and have `#[test]`, but those
+    // are omitted since the example runs without the rust test harness so they
+    // are called by main directly.
+
+    // It is possible to use `MyHeap<T>` directly as the state under test, but
+    // we add this wrapper so we can use a "known correct" container to track
+    // what's supposed to be in the heap.
+    #[derive(Debug, Clone)]
+    struct MyHeapState<T> {
+        heap: MyHeap<T>,
+        // Vec is pretty inefficient for the test here, but we don't mind the
+        // test being slow in exchange for something "obviously correct" to
+        // test against.
+        existing_elements: Vec<T>,
+    }
+
+    // We define a strategy for our starting state. Here, we always start with
+    // an empty heap.
+    fn initial_heap_strategy<T : cmp::Ord + fmt::Debug + Clone>
+        () -> impl Strategy<Value = MyHeapState<T>>
+    {
+        Just(MyHeapState {
+            heap: MyHeap::new(),
+            existing_elements: vec![],
+        })
+    }
+
+    // Now for the meat of the test: our state machine!
+    fn heap_state_machine<T>() -> StateMachine<MyHeapState<T>>
+    where T : cmp::Ord + fmt::Debug + Arbitrary + Clone + 'static
+    { prop_state_machine! {
+        type = MyHeapState<T>;
+
+        // First, we define some invariants. That is, properties that
+        // should always hold at all points between mutation operations.
+        invariant len_equals_number_of_elements(&state) {
+            assert_eq!(state.existing_elements.len(), state.heap.len());
+        }
+
+        invariant is_empty_if_length_zero(&state) {
+            assert_eq!(0 == state.heap.len(), state.heap.is_empty());
+        }
+
+        // There are more complex invariants that could be checked, such as
+        // examining whether it still has the heap property. Here, we rely
+        // on the fact that other properties will break if the internal
+        // invariants are omitted, for the sake of being concise.
+
+        // Now for our mutators. We only defined two mutation operations on
+        // our heap so far.
+
+        // push gets extra weight so that we tend to build larger heaps rather
+        // than keeping tiny ones.
+        #[weight = 2]
+        fn push(&mut state, value: T) {
+            // Add it to the heap...
+            state.heap.push(value.clone());
+            // ... and also the Vec where we track what the heap contains.
+            state.existing_elements.push(value);
+        }
+
+        fn pop(&mut state) {
+            let was_empty = state.heap.is_empty();
+
+            match state.heap.pop() {
+                // If None is returned, the heap must have been empty
+                // before the operation.
+                None => assert!(was_empty),
+
+                Some(value) => {
+                    assert!(!was_empty);
+
+                    // The heap must not contain any value which was
+                    // greater than the "maximum" we were just given.
+                    for in_heap in state.heap.iter() {
+                        assert!(value >= *in_heap,
+                                "Popped value {:?}, which was less \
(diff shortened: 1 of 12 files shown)


examples/state_machine_v1.rs
```diff
@@ -0,0 +1,201 @@
...
+/// https://doc.rust-lang.org/stable/std/collections/struct.BinaryHeap.html,
+/// except slow and buggy.
+#[derive(Clone, Debug)]
+pub struct MyHeap<T> {
+    data: Vec<T>,
+}
+
+impl<T : cmp::Ord> MyHeap<T> {
+    pub fn new() -> Self {
+        MyHeap { data: vec![] }
```
Centril: `Self { .. }`

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+Let's kick off by defining our own implementation of a max-heap, like
+`BinaryHeap` from `std`.
+
+```rust
+# use std::cmp;
+#[derive(Clone, Debug)]
+pub struct MyHeap<T> { data: Vec<T> }
+
+impl<T : cmp::Ord> MyHeap<T> {
+    pub fn new() -> Self { MyHeap { data: vec![] } }
```
Centril: `Self { data: .. }`

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+    }
+
+    invariant is_empty_if_length_zero(&state) {
+        assert_eq!(0 == state.heap.len(), state.heap.is_empty());
+    }
+```
+
+Each invariant defines a function which is run against the state. All
+invariants are tested after each mutation. An invariant function may panic or
+return `Err` to fail the test; otherwise, the test is allowed to succeed.
```
Centril: We might want to consider handling `bool` as a return type in some way throughout proptest eventually.
Some inspiration: https://doc.rust-lang.org/nightly/std/process/trait.Termination.html

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+    ) {
+        // The return value is `MyHeapState<i32>` representing the final
+        // state, so we could do some additional checks with it if we
+        // wanted to.
+        test_case.run()?;
+    }
+}
+```
+
+(There currently is no syntax sugar for the above, but there may be in the
```
Centril: Haha you thought of my comment from before :)

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+     $($rest:tt)*) => {
+        #[allow(unused_mut)]
+        let mut attrs = $crate::state_machine::helper::OpAttrs::DEFAULT;
+        $(attrs.$attr = $attrv;)*
+        $ops.push(
+            (attrs.weight,
+             $crate::strategy::Strategy::boxed(
+                 $crate::strategy::LazyJust::new(
+                     || $crate::state_machine::Mutator::boxed(
+                         $crate::state_machine::DirectMutator::new(
```
Centril: These series of calls could perhaps add some pressure on compile times;
Consider creating a hidden exported helper function for this?

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+///
+/// `StateMachine` values are simultaneously a description of mutations and
+/// invariants of that state, as well as a test that the state under test
+/// upholds its expected behaviours.
+///
+/// Instances of this struct are usually constructed by the
+/// `proptest_state_machine!` macro. Please see the documentation of that macro
+/// for examples of typical usage.
+#[must_use = "StateMachine does nothing unless used"]
+pub struct StateMachine<S : fmt::Debug> {
```
Centril: A bound `S: 'static` is implied here because of WF (well-formedness) conditions on the fields...

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+    mutation_strategy: BoxedStrategy<Box<dyn Mutator<S>>>,
```
Centril: + `BoxedStrategy<?A>` requires `?A: 'static`
+ `?A = Box<dyn 'static + Mutator<S>>`
+ This requires `S: 'static`

Proof:

```rust
use std::fmt;

trait Mutator<S> {}

impl<'a, S> Mutator<S> for &'a str {}

pub struct StateMachine<S: fmt::Debug> {
    ms: Box<dyn Mutator<S>>,
}

fn main() {
    let string: String = "foobar".into();
    let borrow: &str = &string;
    let ms: Box<dyn Mutator<&str>> = Box::new(borrow);
    // Remove the line below and it is OK!
    StateMachine { ms };
}
```

Results in:

```rust
   Compiling playground v0.0.1 (file:///playground)
error[E0597]: `string` does not live long enough
  --> src/main.rs:13:25
   |
13 |     let borrow: &str = &string;
   |                         ^^^^^^ borrowed value does not live long enough
...
17 | }
   | - borrowed value only lives until here
   |
   = note: borrowed value must be valid for the static lifetime...
```

If we change to the following it will work fine; but that might cause problems elsewhere:

```rust
pub struct StateMachine<'a, S: fmt::Debug> {
    ms: Box<dyn Mutator<S> + 'a>,
}

File: src/state_machine.rs [hunk 8]
```diff
@@ -0,0 +1,1062 @@
...
+        StateMachine {
+            mutation_strategy: self.mutation_strategy.clone(),
+            invariants: Arc::clone(&self.invariants),
+        }
+    }
+}
+
+// TODO: Why does rustc think `S : 'static` is necessary? It seems to arise
+// from `Strategy<Value = Box<dyn Mutator<S>>> + 'static` somehow implying
+// `S : 'static`.
```
Centril: See above; `S: 'static` arises immediately from `StateMachine<S>`.

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+    /// against the state value type.
+    ///
+    /// `invariants` is a function which is run immediately after each
+    /// mutation, to check for properties of the state that are expected to
+    /// hold at all times. The function takes `&mut S` to allow `delegate` to
+    /// use a single mapping function. Invariants should not actually modify
+    /// the value.
+    pub fn new(mutation_strategy: impl Strategy<Value = Box<dyn Mutator<S>>> + 'static,
+               invariants: Arc<dyn Fn (&mut S) -> TestCaseResult>) -> Self {
+        StateMachine {
```
Centril: `Self { .. }`

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+
+    /// Run this test case.
+    ///
+    /// Returns `Ok` with the final internal state if successful. On failure,
+    /// returns `Err` with the appropriate error, or panics if user code
+    /// triggered a panic.
+    pub fn run(mut self) -> Result<S, TestCaseError> {
+        if self.trace {
+            println!("initial state: {:?}", self.state);
+            println!("check initial invariants");
```
Centril: A thought for later: It might be nice to change from using `println` to use some more principled logging facility that users can control. I think the quickcheck crate does this. But this is out of scope for this PR.

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+        Ok(self.state)
+    }
+}
+
+/// A single-use callback that can mutate a particular type of `StateMachine`.
+pub trait Mutator<S> : fmt::Debug {
+    /// Mutate `state`.
+    ///
+    /// This effectively consumes `self`, but it is passed as a mutable
+    /// reference so that this trait is object-safe and so that the mutator can
```
Centril: Add a test that asserts object safety.

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+/// state, then runs another mutator against the interior value.
+pub struct MapMutator<
+    S, C,
+    A : fmt::Debug,
+    F : FnOnce (&mut S, A) -> Result<&mut C, TestCaseError>
+> {
+    function: Option<F>,
+    arguments: Option<A>,
+    formatted_arguments: Option<Cow<'static, str>>,
+    name: &'static str,
```
Centril: These fields are in common from the previous struct... Reuse?

src/state_machine.rs
```diff
@@ -0,0 +1,1062 @@
...
+
+impl<S, C, A : fmt::Debug,
+     F : FnOnce (&mut S, A) -> Result<&mut C, TestCaseError>>
+Mutator<S> for MapMutator<S, C, A, F> {
+    fn mutate(&mut self, state: &mut S) -> TestCaseResult {
+        let fun = self.function.take().expect("Mutator applied more than once");
+        let args = self.arguments.take().expect("Mutator applied more than once");
+        if self.formatted_arguments.is_none() {
+            self.formatted_arguments = Some(Cow::Owned(format!("{:?}", args)));
+        }
```
Centril: This bit is identical from `DirectMutator`; refactor?

Centril:
Additional thoughts:

+ The base of this is quite nice.

+ I wonder if sugar should be added for preconditions of mutators (they can easily be added as `if`s in each mutator but sugar could be nice...). This would model the fact that in state machines, after visiting some state nodes, a state machine may enter a phase which causes it to never enter those previous state nodes.

+ Sugar for post conditions could also be nice. You could add optional `require` and `ensure` clauses to each mutator to do this and the previous point.

+ Is "mutation" the right word? Other words to consider "command", "action". Tho perhaps "mutation" is the rustic word :)

+ The major question for me is whether and how this can be used for tests involving concurrent situations such as async or involving mutexes and so on... In particular, can we do the things in http://scalacheck.org/files/scaladays2014/#27 and the following slides?

   In particular, there's the question of symbolic vs. dynamic execution (see https://medium.com/@jlouis666/breaking-erlang-maps-2-362730a91400 for a discussion). In the blog post, they note that they generate the command sequence (and I assume that would include delegated bits too) before invoking any of the commands.

    From reading the code in this PR I'm not 100% sure what is done wrt. symbolic / dynamic execution, but I think that the command structure, including delegates are generated before any mutations actually occur (which is good). Could you say a few words about this? I think it would be good to include in the documentation or at least the code a bit about the high level logic of how the execution model works possibly including some diagrams / pictures or ascii art as in the medium blog post.

   I think it would be good to have some test cases for concurrent situations.

Eh2406:
I was working on a data structure in Cargos internals and remembered this work. How is it going? What are the next steps?

AltSysrq:
To be honest, I haven't really done anything here since I made the PR.  IIRC I ran into some fundamental flaw in the way this is designed and put it on the backburner, but I've forgotten what that was now.

zmitchell:
I'd be interested in continuing work on this if someone is able to mentor me

matthew-russo:
This got deleted when I switched from branch `master` to `main`. We have added state-machine support via https://crates.io/crates/proptest-state-machine so I will leave this closed

PR #100, opened 2018-10-31
Bump syn => 0.15.17

Centril:
See title ;)


Centril:
:) 

Btw, Travis CI now supports, https://blog.travis-ci.com/2018-10-11-windows-early-release, testing on Windows so maybe we should check that out (since AppVeyor is slow...)?

AltSysrq:
Cool, thanks for pointing that out.

PR #103, opened 2018-11-01
Implement #[proptest(regex = "...")]

Centril:
This PR implements `#[proptest(regex = "...")]`.

+ To do so cleanly, the trait `StrategyFromRegex` is introduced into `proptest::string` but is hidden and semver exempt.
+ The PR improves some of the error messages.
+ New error messages E0034 (reports malformed `regex` attribute modifiers), E0035 (reports `params` and `regex` used on the same field together which is nonsense since the `params` can't be used inside the regex string...).
+ Some minor refactoring is done.
+ Some missing test coverage is made up for.


AltSysrq:
Thanks again!

PR #104, opened 2018-11-03
Proptest-derive book

AltSysrq:
Adds the basic mdbook skeleton. This PR will eventually include the full book for `proptest-derive`.

There's some stubs for the `proptest` section of the book, but that will not be filled further in the context of this PR.


book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+    V2(!, !),
+}
+```
+
+## E0006
+
+This error occurs if `#[derive(Arbitrary)]` is used on an enum where all
+inhabited variants are marked with `#[proptest(skip)]`. In other words,
+proptest is forbidden from generating any of the enum's variants, and thus the
+enum itself is unconstructable.
```
Centril:
```suggestion
enum itself cannot be generated.
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+
+Example:
+
+```rust
+#[derive(Debug, Arbitrary)]
+#[proptest(value = "MyStruct(42)")]
+struct MyStruct(u32);
+```
+
+These attributes are only applicable to fields.
```
Centril:
```suggestion
These attributes are only applicable to fields and enum variants.
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+
+```rust
+#[derive(Debug, Arbitrary)]
+struct WidgetContainer {
+    desired_widget_count: usize,
+    #[proptest(value = "vec![]")] // Always generate an empty widget vec
+    widgets: Vec<Widget>,
+}
+```
+
```
Centril:
```suggestion

The only things that can be skipped are enum variants. For example:

```rust
#[derive(Debug, Arbitrary)]
enum Foo {
    NormalVariant(u8),
    #[proptest(skip)]
    MySpecialVariant(String),
}
```

You'll need to add ``` to this because GitHub won't let me add them...

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+#[derive(Debug, Arbitrary)]
+struct Foo {
+    a: u32,
+    #[proptest(weight = 42)]
+    b: u32,
+}
+```
+
+The `weight` attribute only is sensible where proptest has a choice between
+multiple items, such as enum variants. In contrast, with struct fields proptest
```
Centril:
```suggestion
multiple items, that is: enum variants. In contrast, with struct fields proptest
```

"such as" suggests that there may be more but it enum variants is the only thing literally.

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+This error occurs if `#[proptest(filter = "expr")]` is set on an item, but the
+item containing it specifies a direct way to generate the whole value, which
+would thus occur without consulting the filter.
+
+Example:
+
+```rust
+#[derive(Debug, Arbitrary)]
+enum Foo {
+    #[proptest(value = "Foo::V1(42)")]
```
Centril:
```suggestion
    #[proptest(value = "Foo::Bar(42)")]
```

It would be nice to have some "funner" names... `Bar` isn't that much better, but well, it fits :)

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+    V1 {
```
Centril:
```suggestion
    Bar {
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+        #[proptest(filter = "is_even")]
+        field: u32,
+    },
+    // ...
+}
+```
+
+In this example, the entire `V1` variant specifies how to generate itself
```
Centril:
```suggestion
In this example, the entire `Bar` variant specifies how to generate itself
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+
+This error would occur if an outer attribute of the form `#![proptest(..)]`
+were applied to something underneath a `#[derive(Arbitrary)]`.
+
+As of Rust 1.30.0, there are no known ways to produce this error since the Rust
+compiler will reject the attribute first.
+
+## E0014
+
+This error occurs if a bare `#[proptest]` attribute is applied to anything.
```
Centril:
```suggestion
This error occurs if a bare `#[proptest]` attribute is applied to anything,
since it has no meaningful content.
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+    #[proptest]
+    field: u8,
+}
+```
+
+The only legal use of the attribute is the form `#[proptest(..)]`.
+
+## E0015
+
+This error occurs if an attribute of the form `#[proptest = value]` is
```
Centril: Aside: I've wondered if we should give this attribute form meaning...
Maybe it should be synonymous with `#[proptest(strategy = expr)]` but that is perhaps not super clear;
On enum variants, I would also expect this to be for weights...

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+## E0024
+
+This error happens if an invalid `#[proptest ..]` attribute is applied using a
+syntax the `proptest-derive` crate is not prepared to handle.
+
+Exactly what conditions can produce this error vary by Rust version.
+
+## E0025
+
+This error happens if both `#[proptest(strategy = "expr")]` and
```
Centril:
```suggestion
This error happens if more than one of `#[proptest(strategy = "expr")]`,
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+`#[proptest(value = "expr")]` are applied to the same item.
```
Centril:
```suggestion
`#[proptest(value = "expr")]`, or `#[proptest(regex = "string")]` are applied to the same item.
```

I'm assuming we land https://github.com/AltSysrq/proptest/pull/103 here. :)

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+
+```rust
+#[derive(Debug, Arbitrary)]
+struct Foo {
+    #[proptest(value = "42", strategy = "Just(56)")]
+    bar: u32,
+}
+```
+
+Both of these modifiers completely describe how to generate the value, so they
```
Centril:
```suggestion
Any one of these modifiers completely describe how to generate the value, so not
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+cannot both be applied to the same thing. One or the other must be chosen
```
Centril:
```suggestion
more than one can be applied to the same thing. One or the other must be chosen
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+```
+
+There are a few different ways to get this error:
+
+- Pass nothing at all. E.g., `#[proptest(value)]`.
+
+- Use another illegal form. E.g., `#[proptest(value("a", "b"))]`.
+
+- Pass a string expression which is not valid Rust syntax, as in the above
+  example.
```
Centril:
```suggestion
  example. As with E0023, what is valid Rust syntax is determined by the `syn` crate.
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+
+This error happens if both `#[proptest(strategy = "expr")]` and
+`#[proptest(value = "expr")]` are applied to the same item.
+
+Example:
+
+```rust
+#[derive(Debug, Arbitrary)]
+struct Foo {
+    #[proptest(value = "42", strategy = "Just(56)")]
```
Centril:
```suggestion
    #[proptest(value = "42", strategy = "Foo { bar: 56 }")]
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+```
+
+Both of these modifiers completely describe how to generate the value, so they
+cannot both be applied to the same thing. One or the other must be chosen
+depending on the desired effect.
+
+## E0026
+
+This error happens if an invalid form of `#[proptest(strategy ..)]` or
+`#[proptest(value..)]` is used.
```
Centril:
```suggestion
`#[proptest(value ..)]` is used.
```

( pick only one of these suggestions... =P )

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+}
+```
+
+Both of these modifiers completely describe how to generate the value, so they
+cannot both be applied to the same thing. One or the other must be chosen
+depending on the desired effect.
+
+## E0026
+
+This error happens if an invalid form of `#[proptest(strategy ..)]` or
```
Centril:
```suggestion
This error happens if an invalid form of `#[proptest(strategy..)]` or
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+```
+
+There are a few different ways to get this error:
+
+- Pass nothing at all. E.g., `#[proptest(filter)]`.
+
+- Use another illegal form. E.g., `#[proptest(filter("a", "b"))]`.
+
+- Pass a string expression which is not valid Rust syntax, as in the above
+  example.
```
Centril:
```suggestion
  example. As with E0023, what is valid Rust syntax is determined by the `syn` crate.
```

book/src/proptest-derive/errors.md
```diff
@@ -0,0 +1,631 @@
...
+    _bar: PhantomData<T>,
+}
+```
+
+The only valid form for the modifier is `#[proptest(no_bound)]`.
+
+## E0033
+
+This error occurs if the sum of the weights on the variants of an enum overflow
+a `u32`.
```
Centril:
```suggestion
an `u32`.
```

book/src/intro.md
```diff
@@ -0,0 +1,39 @@
+# Introduction
+
+Proptest is a property testing framework (i.e., the QuickCheck family)
+inspired by the [Hypothesis](http://hypothesis.works/) framework for
+Python. It allows to test that certain properties of your code hold for
```
Centril:
```suggestion
Python. It allows you to test that certain properties of your code hold for
```

book/src/intro.md
```diff
@@ -0,0 +1,39 @@
...
+crates which enables reusing those implementations with proptest.
+
+## Status of this crate
+
+The majority of the functionality offered by proptest is in active use and
+is known to work well.
+
+The API is unlikely to see drastic breaking changes, but there may still be
+minor breaking changes here and there, particularly when "impl Trait"
+becomes stable and after the upcoming redesign of the `rand` crate.
```
Centril: This bit has mostly happened?

I think maybe we should point to GATs instead now...

book/src/proptest-derive/pd/index.md
```diff
@@ -0,0 +1,9 @@
+# The `proptest-derive` crate
```
Centril: Hmm this file seems to be an unused duplicate?

book/src/proptest-derive/errors.md
AltSysrq: This seems a bit extraneous; it's not relevant to the user's problem because they're clearly not working with an enum, and switching to using an enum is not a sensible solution.

Centril: Hmm; I suppose -- but showing what the attribute is used for is nice, so maybe link to other parts of the book (once written) that shows the use of the attribute?

book/src/proptest-derive/errors.md
AltSysrq: That syntax would feel unclear to me. Let's keep it explicit for now, we can always add some meaning to it later.

Centril: Yeah I agree 👍

book/src/proptest-derive/errors.md
AltSysrq: How is `Foo { bar: 56 }` a valid argument for `strategy` here?

Centril: Ignore me; I was being silly here!

book/src/proptest-derive/pd/index.md
AltSysrq: I plan to deduplicate it once I start actually working on the proptest section. Roughly, my plan is to move most of the exposition in `lib.rs` to the book, and then generate the readme from sections of the book.

book/src/proptest-derive/errors.md
AltSysrq: How do you pronounce it? I say "you thirty two" which definitely takes "a".

book/src/proptest-derive/errors.md
AltSysrq: Yeah, that sounds like a good plan to me.

book/src/proptest-derive/pd/index.md
Centril: Ah :)

book/src/proptest-derive/errors.md
AltSysrq: I've applied this to most of the examples.

book/src/proptest-derive/errors.md
AltSysrq: I moved the discussion about workarounds to a separate section at the bottom that everything just links to.

book/src/proptest-derive/errors.md
Centril: Interesting; I say it the same way so you are right here. :)

book/src/proptest-derive/errors.md
Centril: Yeah; really like the improvements!

book/src/proptest-derive/errors.md
Centril: Saw that; love it!

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+For brevity, modifiers are sometimes referenced by name alone; e.g., "the
+`weight` modifier" refers to `#[proptest(weight = nn)]` and not some
+freestanding `#[weight]` attribute.
+
+## `filter`
+
+Form: `#[proptest(filter = F)]` or `#[proptest(filter(F))]` where `F` is either
+a bare identifier or a Rust expression in a string. In either case, the
+parameter must evaluate to something which is `Fn (&T) -> bool`, where `T` is
```
Centril:
```suggestion
parameter must evaluate to something which is `Fn(&T) -> bool`, where `T` is
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+attributes with one modifier each.
+
+For brevity, modifiers are sometimes referenced by name alone; e.g., "the
+`weight` modifier" refers to `#[proptest(weight = nn)]` and not some
+freestanding `#[weight]` attribute.
+
+## `filter`
+
+Form: `#[proptest(filter = F)]` or `#[proptest(filter(F))]` where `F` is either
+a bare identifier or a Rust expression in a string. In either case, the
```
Centril: Clarify that "bare identifier" refers to a function (or const/static also works...)?

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+    end: i32,
+}
+```
+
+As mentioned above, filtering should be avoided when it is reasonably possible
+to express a non-filtering strategy that achieves the same effect. For example:
+
+```rust
+#[derive(Debug, Arbitrary)]
+struct BadExapmle {
```
Centril:
```suggestion
struct BadExample {
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+    // Don't do this! Your tests will run more slowly and shrinking won't work
+    // properly.
+    #[proptest(filter = "|x| x % 2 == 0")]
```
Centril: Maybe extract this out to an `is_even` function to demonstrate how that works?

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+}
+```
+
+[`prop_filter`]: https://altsysrq.github.io/rustdoc/proptest/latest/proptest/strategy/trait.Strategy.html#method.prop_filter
+
+## `no_bound`
+
+Form: `#[proptest(no_bound)]`
+
+Usable on: generic type declarations
```
Centril:
```suggestion
Usable on: generic type definitions and type parameters
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+Normally, when `#[derive(Arbitrary)]` is applied to an item with generic type
+arguments, every type argument is required to `impl Arbitrary`. For example,
```
Centril:
```suggestion
parameters, every type parameter is required to `impl Arbitrary`. For example,
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+Placing `#[proptest(no_bound)]` on a generic type argument suppresses this. For
+example, the following removes the extra `where T: Arbitrary`:
+
+```rust
+#[derive(Arbitrary)]
+struct MyStruct<#[proptest(no_bound)] T> { /* ... */ }
+```
+
+In order for `no_bound` to be useful, one of two things needs to be true:
```
Centril:
```suggestion
For `no_bound` to be useful, one of two things needs to be true:
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+impl<T> Arbitrary for MyStruct<T> where T: Arbitrary { /* ... */ }
+```
+
+Placing `#[proptest(no_bound)]` on a generic type argument suppresses this. For
+example, the following removes the extra `where T: Arbitrary`:
+
+```rust
+#[derive(Arbitrary)]
+struct MyStruct<#[proptest(no_bound)] T> { /* ... */ }
+```
```
Centril: I'd also include an example:


```rust
#[derive(Arbitrary)]
#[proptest(no_bound)]
struct MyStruct<T, U> { /* ... */ }
```

since that short-hand also exists.

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+- Other modifiers (such as `value`) are added to things referencing the generic
+  type that remove proptest's responsibility for generating them.
+
+Example of the above:
+
+```rust
+#[derive(Debug, Arbitrary)]
+enum MyEnum<#[proptest(no_bound)] T> {
+    // OK -- no T needs to be constructed
+    Phantom(PhantomData<T>),
```
Centril: The macro recognizes `PhantomData<T>` ("magic", see `if util::is_phantom_data(path) { return; }`) and will not add a bound for that. See tests in `tests/phantom.rs`.

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+    // OK -- we're manually taking responsibility for generating the value
+    #[proptest(value = "MyEnum::WithValue(None)")]
+    WithValue(Option<T>),
```
Centril: A consequence of taking responsibility is that `T: Arbitrary` goes away as well.

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+    // OK -- we tell proptest not to generate this case
+    #[proptest(skip)]
+    Skipped(T),
```
Centril: A consequence of skipping is that `T: Arbitrary` won't be added also.

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+    // ERROR -- proptest would need to generate a T here, but cannot because
+    // we lack the `T: Arbitrary` bound due to having used `no_bound`.
+    Error(T),
```
Centril: But this is correct; In general, when you make the macro use `any` or `any_with` then `T: Arbitrary` will be added... You can see this by looking for `.mark_uses(` and some of the logic inside `UseTracker::add_bounds`.

This also applies when you use some associated type... If we we write (`tests/assoc.rs`):
```rust
#[derive(Debug, Arbitrary)]
struct T4<Tyvar: FuncB> {
    field: Tyvar::OutB,
}
```
then we require `<Tyvar as FuncB>::OutB: Arbitrary` but not `Tyvar: Arbitrary`.

The following situation would require `no_bound` tho:

```rust
#[derive(Debug, Arbitrary)]
struct Hide<T>(PhantomData<T>);

#[derive(Debug, Arbitrary)]
struct Stuff<T>(Hide<T>);
// We have no idea that `Arbitrary for Hide<T>` doesn't require `T: Arbitrary`
// so we use `no_bound` to get rid of the extra bound.
```

I should probably add a new round of tests to ensure the semantics are as expected and more robust in the face of changes. Some documentation improvements in the code itself are also probably necessary on my part.

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+`#[proptest(..)]`, where the content between the parentheses follows the normal
+Rust attribute syntax.
+
+Each modifier within the parentheses is independent, in that putting two
+modifiers in the same attribute is equivalent to having two `#[proptest(..)]`
+attributes with one modifier each.
+
+For brevity, modifiers are sometimes referenced by name alone; e.g., "the
+`weight` modifier" refers to `#[proptest(weight = nn)]` and not some
+freestanding `#[weight]` attribute.
```
Centril: Hmm... I think `weight = W` works, but a reader might now have read this bit and is jumping around and so `#[weight]` could be confusing...

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+    // ERROR -- proptest would need to generate a T here, but cannot because
+    // we lack the `T: Arbitrary` bound due to having used `no_bound`.
+    Error(T),
+}
+```
+
+## `no_params`
+
+Form: `#[no_params]`
```
Centril:
```suggestion
Form: `#[proptest(no_params)]`
```

I'd write it out fully; saves some confusion in some cases imo.

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+    #[proptest(strategy = "params.0 ..= params.1")]
+    desired_widget_count: usize,
+    // ...
+}
+
+// ...
+
+proptest! {
+    #[test]
+    fn test_something(wc in any::<WidgetCollection>(WidgetRange(10, 20))) {
```
Centril:
```suggestion
    fn test_something(wc in any_with::<WidgetCollection>(WidgetRange(10, 20))) {
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+## `skip`
+
+Form: `#[proptest(skip)]`
+
+Usable on: enum variants
+
+Annotating an enum variant with `#[proptest(skip)]` prevents proptest from
+generating that particular variant. This is useful when there is no sensible
+way to generate the variant.
```
Centril:
```suggestion
way to generate the variant or when you want to temporarily stop generating some variant during development.
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+```rust
+#[derive(Debug, Arbitrary)]
+enum Token {
+    Delimitation {
+        // This field is still generated via Arbitrary
+        delimiter: Delimiter,
+
+        // But for this field we use a custom strategy
+        #[proptest(strategy = "1..10")]
+        count: u32,
```
Centril: Maybe add an example of a bare identifier as an extra field?

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+
+Form: `#[proptest(weight = W)]` or `#[proptest(weight(W))]`, where `W` is an
+expression evaluating to a `u32`. `weight` may also be abbreviated to `w`, as
+in `#[proptest(w = W)]`.
+
+Usable on: enum variants
+
+The `weight` modifier determines how likely proptest is to generate a
+particular enum variant. Weights are relative to each other; for example, a
+`weight=3` variant is 50% more likely to be generated than a `weight=2` variant
```
Centril:
```suggestion
`weight = 3` variant is 50% more likely to be generated than a `weight = 2` variant
```

book/src/proptest-derive/modifiers.md
```diff
@@ -0,0 +1,370 @@
...
+and three times as likely to be generated as a `weight=1` variant.
```
Centril:
```suggestion
and three times as likely to be generated as a `weight = 1` variant.
```

book/src/proptest-derive/modifiers.md
AltSysrq: Hm, I'm not sure how I feel about that… even the built in derives don't do that.

book/src/proptest-derive/modifiers.md
AltSysrq: I added a separate example to the main usage example.

book/src/proptest-derive/modifiers.md
AltSysrq: I had not noticed that was supported. Added it to the docs.

AltSysrq:
The book could probably do with more expansion, but I'm going to move forward with what I have here in order to get the ball rolling on a release since the docs should be sufficient for people to get started at least.

Centril:
Yeah that sounds good;

I might take a pass over the text after you've merged it and maybe make a PR. Thanks for your hard work. I'll try to get back to proptest-derive in a few days but I've been swamped with language and release team duties. :)

Btw, Rust 1.32 has released, so maybe we should do a 0.9.0 release soon (but there are some things I'd like to go over before that...).

AltSysrq:
Yeah, I was planning on working towards 0.9.0.

PR #106, opened 2018-11-06
[EXPERIMENT] proptest-derive: #[inline] + `no_cfg` Cargo feature gate

Centril:
Changes:

- Formatting fix + rewording in `proptest-derive/Cargo.toml`.

- A default enabled cargo feature `no_cfg` is introduced in `proptest-derive/Cargo.toml`.
   - Unless `no_cfg` is active then `#[cfg(test)]` will be put on implementations (generated by `#[derive(Arbitrary)]`).

   - This entails that you can get the `#[cfg(test)]` behavior with `default-features = false`.

- `#[inline]` is put on `arbitrary_with` generated by `#[derive(Arbitrary)]`. This change was suggested by eddyb; the idea being that we don't do ahead-of-time codegen of the implementations produced.

- Tests for the `no_cfg` behavior above; in particular, we set `default-features = false` and ensure that `Arbitrary` isn't provided.

One drawback to this approach where cargo is used is that if one of the dependencies use `no_cfg` then all of them will; but I don't think this should be that much of a problem; I think it is better than the ergonomics hit you get by having to annotate each `derive(Arbitrary)` location instead.

cc https://github.com/AltSysrq/proptest/issues/102, @Nemo157, @Eh2406


Centril:
The PR is pretty much done and is working; The question now is whether this is the design we want.

Nemo157:
I’ll try and give proptest-derive a spin tonight and see how I like this for my usecase.

AltSysrq:
> if one of the dependencies use no_cfg then all of them will

It does seem like a problem to me, unless I'm misunderstanding something.

Crate `dep-a`:

```toml
[dependencies]
proptest-derive = { version = "xx", default-features = false }

[dev-dependencies]
proptest = "xx"
```

Crate `dep-b`:
```toml
[dependencies]
proptest-derive = "xx"
proptest = "xx"
```

Crate `bin`:
```toml
[dependencies]
dep-a = "yy"
dep-b = "zz"
```

`dep-b` forces the `no_cfg` feature to be on everywhere. But now in `dep-a` `#[derive(Arbitrary)]` tries to reference `proptest` in the generated code, but the `proptest` crate is not available to `dep-a`.

Centril:
Aw! bummer :(
You're right, this creates a big problem for the whole setup.

This problem could be fixed with:
```rust
#[cfg(all(
    any(test, feature = "prop"),
    accessible(::proptest::prelude::Arbitrary)
    // `dep-a` will not have `::proptest::*` accessible so you get no impl
    // and thus also no error.
)]
impl Arbitrary for Foo { ... }
```
but unfortunately we don't have [`accessible(...)`](https://github.com/rust-lang/rfcs/pull/2523) yet.

At this point it might be best to just standardize on `feature = "prop"` and we can perhaps also have `#[proptest(cfg)]` which will add `#[cfg(any(test, feature = "prop"))]` on the impl...?

(We should also retain `#[inline]` from this PR...)

Centril:
I think the new proposal works even for those who want to do:

```rust
#[derive(Debug)]
#[cfg_attr(feature = "prop", derive(Arbitrary))]
struct Foo;
```

as well... so that is good.

cc @Nemo157 what do you think? does that work for you?

PS: I would not name the feature `proptest` tho because that would conflict with the crate `proptest` due to deficiencies in Cargo. `prop` is shorter and doesn't conflict so it should work; it's also fairly unique I think so that's also good.

AltSysrq:
> I would not name the feature `proptest` tho because that would conflict with the crate `proptest`

Is there something I don't know here? I was deliberately using feature == optional crate so that there's less to wire up.

Nemo157:
@AltSysrq the problem is when you have a dependency that has its own optional `proptest` feature. You would want to enable it when a user enables your `proptest` feature, but you can’t override the meaning of features created by optional dependencies.

It may pay to talk to the `serde` team about this, they should have the same issue but they strongly recommend using the name `serde` anyway; maybe they have some ideas of how to improve on this.

Centril:
> Is there something I don't know here? I was deliberately using feature == optional crate so that there's less to wire up.

Oh so someone (crate "alpha") would do:

```toml
name = "alpha"

[dependencies]
proptest = { version = "x.y", optional = true }
```

and then someone else (crate "beta") would do:

```toml
name = "beta"

[dependencies]
# For the direct stuff we want to derive:
proptest = { version = "x.y", optional = true }

[dependencies.alpha]
version = "x.y"
features = ?
#^ How do we make: proptest imply alpha/proptest?
```
https://doc.rust-lang.org/cargo/reference/manifest.html#rules

Arjentix:
Just found this PR in [book](https://proptest-rs.github.io/proptest/proptest-derive/getting-started.html) and there were 5 years since the last update. Is it alive? If not, maybe remove the mention from the book?

PR #112, opened 2018-12-13
remove the nightly requirement for lazy_static on `no_std`

nivkner:
thanks to rust-lang-nursery/lazy-static.rs#130, lazy-static now works on stable with `no_std`

nightly is still required for `alloc` though, so this isn't `no_std` proptest on stable yet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,10 +28,7 @@ default = ["std", "fork", "timeout", "bit-set"]
 # Enables unstable features of Rust.
 unstable = ["rand/i128_support"]
 
-# Required for no_std usage thanks to lazy_static's
-# dependency on the spin crate, which is nightly-only
-# for no_std.
-nightly = ["lazy_static/spin_no_std"]
+nightly = []
 
 # Enables the use of standard-library dependent feature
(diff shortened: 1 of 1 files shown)


Eh2406:
I believe removing a feature is a break-in change.

nivkner:
yeah, I guess it is.
removed the commit, but since the nightly feature was only used for this,
it should be removed for 0.9

AltSysrq:
Thanks for this!

0.9 is going to be the next version anyway, but I think it makes sense to keep the `nightly` feature since we may be adding stuff to it in the future again.

nivkner:
@AltSysrq is there a reason why this PR did not get merged?
since as previously stated I removed the feature removing commit, this only removes its use for lazy_static.

Centril:
@AltSysrq We have an [`unstable`](https://github.com/AltSysrq/proptest/blob/master/Cargo.toml#L29) feature, so we should remove `nightly` in favor of that one (unstable is the standard name in the ecosystem).

AltSysrq:
This has been incorporated into 0.9.0.

PR #113, opened 2018-12-15
chore: rand 0.5 → 0.6

ignatenkobrain:


ignatenkobrain:
Only this warning is remaining:

```rust
warning: use of deprecated item 'rand::seq::sample_iter': use IteratorRandom::choose_multiple instead                                       
   --> src/bits.rs:267:13                                                                                                                   
    |                                                                                                                                       
267 |             rand::seq::sample_iter(runner.rng(), self.bits.iter(), count)                                                             
    |             ^^^^^^^^^^^^^^^^^^^^^^                                                                                                    
    |                                                                                                                                       
    = note: #[warn(deprecated)] on by default                                                                                               
```

Centril:
I think you'll want to get rid of the `i128` feature enabling here: https://github.com/AltSysrq/proptest/blob/master/Cargo.toml#L29 since it AFAIK is no longer needed in rand-0.6

AltSysrq:
Thanks! This change has gone out in 0.9.0.

PR #123, opened 2019-01-21
Expose TestRunner::partial_clone publicly

Lederstrumpf:
`test_runner::runner::partial_clone` is currently public within the proptest crate only. Making it publicly available outside the crate would be beneficial.
It is useful, for instance, when multiple test runners are required within a test, and these are to be deterministically derived from a single persistent failure saved in `proptest-regressions`. 

Does this align with the design philosophy and intended API surface of the library? Otherwise, if better suited, I would propose an optional `TestRng` argument for instantiating a new `Test_runner::runner::TestRunner`, or some equivalent of this.

### DIFF
--- a/src/test_runner/runner.rs
+++ b/src/test_runner/runner.rs
@@ -282,7 +282,7 @@ impl TestRunner {
     /// Create a fresh `TestRunner` with the same config and global counters as
     /// this one, but with local state reset and an independent `Rng` (but
     /// deterministic).
-    pub(crate) fn partial_clone(&mut self) -> Self {
+    pub fn partial_clone(&mut self) -> Self {
         TestRunner {
             config: self.config.clone(),
             successes: 0,


AltSysrq:
The function is non-public currently because the set of fields that are copied vs shared is somewhat arbitrary and I didn't want to commit to keeping compatibility.

> I would propose an optional `TestRng` argument

I prefer this option.

If you don't mind, could you provide some more detail on why you have a test that needs multiple `TestRunner`s though?

AltSysrq:
0.9.1 includes a `new_with_rng` function on `TestRunner`, which can be used in conjunction with `TestRunner::new_rng` to accomplish what you want.

Lederstrumpf:
Thanks - sorry that I lost track of this!

> If you don't mind, could you provide some more detail on why you have a test that needs multiple TestRunners though?

Although my issue is now addressed, I think the answer to this is still relevant information regarding (mis?)usage patterns of the library:

I'm generating a sequence of system states, where, internally, each event takes the prior system state and some generated parameters as an input to generate the new state. So at the highest level, I have a `BoxedStrategy` for a vector of these system states, and the function used to generate these system states uses its own `TestRunner` to generate the parameters of the event leading to the following system state.

So since `TestRunner::run(...)` already takes a mutable reference to `self`, a reference to the TestRunner cannot also be passed to the Strategy.

By using multiple TestRunners, I believe I lose out on shrinking/growing for failing cases, but implementing strategies for these is conceptually hopeless for my scenario anyway.

Look forward to integrating the latest changes!

AltSysrq:
I think I understand. Paraphrasing just to make sure we're both on the same page: You're essentially testing a `fn (State, Operation) -> State`, and the set of allowable `Operation`s depends on the `State` value, so for each iteration you need to make a new `TestRunner` to create the next iteration's `Operation`.

It's definitely a bit gnarly. I assume this is not the case, but in the off chance it is possible for you to fully enumerate the set of `Operation`s at each step, you could have proptest generate [`Index`es](https://altsysrq.github.io/rustdoc/proptest/0.9.1/proptest/sample/struct.Index.html) and select operations that way.

Otherwise, I do have one other thing you could try. As an input for your test, take an argument like this:

```rust
seeds in prop::collection::vec(any::<[u8;32]>().no_shrink(), 1..however_many_operations_you_want)
```

Then, for each iteration, you can use `TestRunner::new_with_rng(ProptestConfig::default(), TestRng::from_seed(RngAlgorithm::ChaCha, seed))`, using successive values of `seeds` for each `seed`. While your `Operation`s won't shrink, it will let proptest attempt to delete `Operation`s entirely. Here is an example of what I'm suggesting: https://gist.github.com/AltSysrq/690421f46dd083914de46fdced554897

Lederstrumpf:
Thanks for your explicit example!

Maybe I should have explicitly stated what I meant by nested test runners:
I do not have a wrapping `proptest!` macro, but perform the runs manually. Reason for this is that I wanted to clone the state of the parent `rng` and pass it to the child (which is now redundant with generating seeds deterministically as part of the input strategy, like in your example).
So in terms of actual test runners, my setup is equivalent to the example you provided (which implicitly has two test runners).

I only terminate the state sequence once some state satisfies a proposition, so in lieu of looping over seeds, I `take_while` from an iterator. I've adapted your example to show the skeleton structure of my use of proptest:
[https://gist.github.com/Lederstrumpf/c5ee2f07da0bffe51b535a6608576886](https://gist.github.com/Lederstrumpf/c5ee2f07da0bffe51b535a6608576886)

I now also realize that using `TestRunner::partial_clone(...)` as in this PR would not have sufficed for a reproducible setup since the cloning of the parent `rng` is necessarily prior to reading in the persistent seeds, which only happens in the body of `TestRunner::run_in_fork(...)`. Happy that the library went the other route!

PR #126, opened 2019-02-09
Fix a type mismatch in the date parser example

dwijnand:
I got this when trying to follow the guide:

    error[E0308]: mismatched types
      --> src/lib.rs:32:24
       |
    32 |             parse_date(s);
       |                        ^
       |                        |
       |                        expected &str, found struct `std::string::String`
       |                        help: consider borrowing here: `&s`
      1 pub fn parse_date(s: &str) -> Option<(u32, u32, u32)> {$                             | 43 //-$
       |
       = note: expected type `&str`
                  found type `std::string::String`

### DIFF
--- a/book/src/proptest/getting-started.md
+++ b/book/src/proptest/getting-started.md
@@ -250,7 +250,7 @@ $ git add proptest-regressions
 ```rust,ignore
 #[test]
 fn test_october_first() {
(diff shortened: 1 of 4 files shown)


AltSysrq:
Good catch! I'll be bringing this in soon.

AltSysrq:
Published in 0.9.1. Thanks!

PR #127, opened 2019-02-09
Avoid a broken relative rustdoc link to Arbitrary

dwijnand:
In [1] the short description of the "arbitrary" module contains a broken
relative link to the Arbitrary trait.  In [2] the link works fine.  This
workarounds the problem by removing the link from the short description
(by pushing it down).

[1] https://altsysrq.github.io/rustdoc/proptest/0.9.0/proptest/index.html#modules
[2] https://altsysrq.github.io/rustdoc/proptest/0.9.0/proptest/arbitrary/index.html

### DIFF
--- a/proptest/src/arbitrary/mod.rs
+++ b/proptest/src/arbitrary/mod.rs
@@ -7,8 +7,10 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! Defines the [`Arbitrary`] trait and related free functions
-//! and type aliases. See the trait for more information.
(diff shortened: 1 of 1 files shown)


dwijnand:
I tried fixing this with https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md, but I couldn't get it to work.

AltSysrq:
_Intra-rustdoc links_ isn't on stable yet, so this is a reasonable enough workaround. I kept forgetting about this issue, so thanks for the PR to fix it! Published in 0.9.1.

PR #128, opened 2019-02-09
Blanket impl for generic parts of FailurePersistence

dwijnand:


dwijnand:
Wow, first CI validation run gave a test failure I couldn't reproduce locally.  The second run gave two test failures.  Have you had trouble with spurious test failures in CI before?

AltSysrq:
This would be a breaking change, and we just had a breaking release a few days ago, so this change would need to wait a while. If you'd like, you could reopen and I'd keep this around for the next time there's a breaking changes release.

> Have you had trouble with spurious test failures in CI before?

Yes. You got unusually unlucky though. Some changes I'm putting in in the next release should fix most of these though.

dwijnand:
Why would it be a breaking change? (Sorry still learning fundamentals)

AltSysrq:
Any `impl`s in other crates before this change were forced to include a definition for all three trait functions this PR affects. After this PR, those functions are no longer on that particular trait, so the compiler would reject the code.

dwijnand:
Yeah, make sense.  I come from the subclassing world, so I assumed pulling up methods would be fine.  Thanks!

I don't have any downstream needs for this, I just noticed the opportunity and wanted to experiment with what was possible, and at that point I might as well pull request it.  Feel free to cherry-pick this change if/when you want it.

PR #129, opened 2019-02-09
A couple of fixups in the book

dwijnand:


AltSysrq:
I definitely like the rewording. This went out along with 0.9.1.

PR #134, opened 2019-03-13
add AsRef<Index> impl for Index

sunshowers:
This is useful for generic code, as the comment says. As a workaround
I've had to define a newtype wrapper around `Index` to achieve the same
goal, so it would be lovely to have this be upstream! 😄

---
This change is [Reviewable](https://reviewable.io/reviews/altsysrq/proptest/134)

### DIFF
--- a/proptest/src/sample.rs
+++ b/proptest/src/sample.rs
@@ -244,6 +244,14 @@ impl Index {
     }
 }
 
+// This impl is handy for generic code over any type that exposes an internal `Index` -- with it,
+// a plain `Index` can be passed in as well.
+impl AsRef<Index> for Index {
+    fn as_ref(&self) -> &Index {
+        self
+    }
+}
+
 mapfn! {
     [] fn UsizeToIndex[](raw: usize) -> Index {
         Index(raw)


sunshowers:
BTW, as someone who's used QuickCheck for a long time, I adore the proptest model and the improvements it brings over QuickCheck. Thank you for writing this! proptest is a delight to work with.

AltSysrq:
I don't have any objections to this PR _per se_, but I do have an honest question: What makes `Index` special? Conceivably this same `impl` could be applied to every type. In `std`, `AsRef` is only used when there are other types that convert to the same thing, but proptest has no other types that could sensibly be `AsRef<Index>`.

sunshowers:
Yeah, good question! I agree that `proptest` itself doesn't have any other types that implement `AsRef<Index>`, but I as a consumer have already written around 5 types so far all of which do implement `AsRef<Index>`. As you're probably aware, the orphan rule would prevent me from doing this in my own crate without using a newtype wrapper.

It's up to you whether you'd like to take it or not, but it would certainly make my life (as someone who's written several thousand lines of proptest-based code) quite a bit easier. I don't know if other types need it -- `Index` has been the only one that's been clear to me so far.

tzemanovic (approved):

PR #136, opened 2019-03-22
Remove requirement on spin crate.

ehuss:
This removes the hard requirement on the `spin_no_std` feature of `lazy_static`.

Cargo uses proptest, but it is also part of the rustc workspace. We would prefer to not include `spin` in rustc, so we need a way to turn it off.

I'm not sure how to test this does what is intended.  This is also just a rough stab on how it could work, any other way to disable it would be fine.

### DIFF
--- a/book/src/proptest/no-std.md
+++ b/book/src/proptest/no-std.md
@@ -15,7 +15,7 @@ default-features = false
 # alloc: Use the `alloc` crate directly. Proptest has a hard requirement on
 # memory allocation, so either this or `std` is needed.
 # unstable: Enable use of nightly-only compiler features.
-features = ["alloc", "unstable"]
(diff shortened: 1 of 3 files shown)


AltSysrq:
I think it should be possible to just eliminate the `lazy_static` in the no-std build altogether — it's only used for caching config derived from the environment, which is completely stubbed out without `std`.

I have some time this weekend, so I'll try to do that then. If not, I'll bring this or something like it in.

The PR builder failure is due to unrelated nightly breakage.

ehuss:
That would be great, thanks! It is very much appreciated!

AltSysrq:
0.9.2 no longer brings the `spin` crate in at all, so closing this PR.

ehuss:
Thanks again, it works great!

PR #144, opened 2019-05-03
Make union value trees lazily generated.

### LINKED ISSUE #143: Union value tree creation has quadratic time complexity in the number of variants
We've been running into some performance issues with large enums and proptest. I did some digging in and it looks like value tree creation is quadratic time in the number of variants involved. That's because value trees are also generated for all previous variants.

For example, for a union with variants 0 through 20, if variant 10 is picked, value trees are generated for all variants 0 through 9 (both inclusive) as well.

Ideally, the value tree generation for variants 0 through 9 would be deferred to shrinking time. I looked at a few approaches on how to do this, and it seems like all of them would require a breaking API change and some of them would require new Rust features:

* Stash away the `&mut TestRunner` into a hypothetical `LazyValueTree` structure. This would require the value tree to carry around a lifetime parameter, and would also require GATs for the `Strategy` trait's `Tree` associated type.
* Do the same as above, except avoid the lifetime parameter by using `Rc` or `Arc` for the test runner. Not great.
* Use a lazy structure, but pass the test runner into `simplify` and `complicate` and use that to initialize the child value trees lazily. This seems like the simplest way forward but it is a major API change and I'm not sure what the other consequences here are.

What do you think?

sunshowers:
I wanted to put this up to get some early opinions. Note that there's the one API change that happens: the
type param for `UnionValueTree` is no longer the tree -- it's the strategy. I don't think this should
affect any actual code, but wanted to flag this just in case.

I'll do lazy tuples as well if this looks good.

Closes #143.


proptest/src/strategy/unions.rs
```diff
@@ -169,8 +173,121 @@ macro_rules! union_value_tree_body {
     }
 }
 
-impl<T : ValueTree> ValueTree for UnionValueTree<T> {
-    union_value_tree_body!(T::Value, access_vec);
+/// `ValueTree corresponding to `Union`.
+pub struct UnionValueTree<T : Strategy> {
+    options: Vec<LazyValueTree<T>>,
+    // This struct maintains the invariant that between function calls,
+    // `pick`, `min_pick` and `prev_pick` (if Some) always point to initialized
```
sunshowers: TODO for me, this comment is wrong, there are no invariants on min_pick.

proptest/src/strategy/lazy.rs
```diff
@@ -0,0 +1,161 @@
...
+            }
+            LazyValueTreeState::Uninitialized { .. } => {
+                write!(f, "Uninitialized")
+            }
+            LazyValueTreeState::Failed => {
+                write!(f, "Failed")
+            }
+        }
+    }
+}
```
sunshowers: TODO for me: add tests

proptest/src/strategy/lazy.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub(crate) fn new_initialized(value_tree: S::Tree) -> Self {
+        Self { state: LazyValueTreeState::Initialized(value_tree) }
+    }
+
+    /// Returns a reference to the inner value tree if initialized.
+    pub(crate) fn as_inner(&self) -> Option<&S::Tree> {
+        match &self.state {
+            LazyValueTreeState::Initialized(v) => Some(v),
+            LazyValueTreeState::Uninitialized { .. }
+            | LazyValueTreeState::Failed => None,
```
Centril: Unless you want to intentionally guard against additions to `LazyValueTreeState` you can just use `_ => None`.
(Applies throughout in this file)

sunshowers: Yeah :) I tend to not use _ because I've run into issues with additions to enums in the past.

proptest/src/strategy/lazy.rs
```diff
@@ -0,0 +1,161 @@
+//-
+// Copyright 2017 Jason Lingle
```
AltSysrq:
```suggestion
// Copyright 2019 The proptest developers
```

proptest/src/strategy/lazy.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl<S : Strategy> Clone for LazyValueTree<S> where S::Tree : Clone {
+    fn clone(&self) -> Self {
+        Self {
+            state: self.state.clone(),
+        }
+    }
+}
+
+// The debug impls here do not require `S` to be Debug to maintain backwards
```
AltSysrq: `Strategy` already implies `Debug`, actually. (Though I don't think that fact would imply any changes to this code.)

proptest/src/strategy/unions.rs
```diff
@@ -169,8 +173,121 @@ macro_rules! union_value_tree_body {
...
+                let initialized = $access!([mut] opt = self, self.pick, {
+                    opt.maybe_init();
+                    opt.is_initialized()
+                });
+                if initialized {
+                    return true;
+                }
+            }
+
+            false
```
AltSysrq: Unless I'm misreading the code, if we get here, `self.pick` is an index to an uninitialised (failed) entry, which breaks `current()`.

sunshowers: Yeah you're correct! I'll fix it with a temp variable and see if I can detect it with tests (think I need some tests with union variants that fail.)

sunshowers:
This is ready for review now! Some notes:
* The overall PR consists of several commits that build on top of each other. The commits are all self-consistent (and should compile at each step) so it may be easier to review each commit separately.
* I've included benchmarks and tests. The benchmarks show an 80-97% improvement in perf for large enums.
* I *believe* this doesn't have any API breakages so can be published as a minor update, but I think that's somewhat debatable. Please let me know if you disagree :)
* I'm not totally sure if I've gotten the non-`std` stuff correct or even what the constraints are, so feedback about that would be greatly appreciated!

Thanks again for proptest.

sunshowers:
Ah, looks like the doctests are failing because of the `std_facade` stuff -- did I get it wrong? https://travis-ci.org/AltSysrq/proptest/jobs/529058312#L2058

AltSysrq:
It looks like the macros currently just assume that `std` is available and somehow none of the `no_std` users have complained.

We could probably make `std_facade` `#[doc(hidden)] pub`, but I'd say for now you should just use `std` directly and I'll fix the macros later.

As for the more general case of `std` vs `no_std`, my usual approach when developing locally is:
- `cargo test` succeeds on stable
- `cargo build --no-default-features --features 'alloc unstable'` succeeds on nightly

The Travis build tests a number of other permutations for certain edge cases but I wouldn't expect you to run into any of those here.

AltSysrq:
Huh, I had forgotten that proptest actually uses the macros internally now. I guess you can go ahead and make `std_facade` accessible.

I generally prefer to do the actual version bumps right before publishing so that the version in the repo matches what's in crates.io, so I'd appreciate it if you removed the version number changes for now. In the changelogs, just change the explicit version to "Unreleased".

sunshowers:
Should `proptest-derive` also use the std facade?

AltSysrq:
The generated code from `proptest-derive` probably should use the std facade. But currently we don't have anything testing that the generated code works with no-std so I wouldn't worry about it too much here.

sunshowers:
Thanks! I won't make any changes then. Anything else I can do to help with the review?

sunshowers:
Hey, just wanted to check in on this :) is there any way I can help?

AltSysrq:
Sorry about the delay. I was going to take care of this last Saturday but (and I know this'll sound like a "dog ate my  homework" level of excuse) but a lightning strike took out my main dev machine and a bunch of other stuff. I should have things mostly back up tomorrow so I should be able to take care of this then.

sunshowers:
Thanks so much! :)

AltSysrq:
These changes have been published in version 0.9.4 and proptest-derive version 0.1.2.

sunshowers:
Thanks again :heart:

PR #151, opened 2019-06-24
Support deriving Arbitrary for structs with more than 10 fields

alessandrod:
This change makes it possible to implement Arbitrary for structs having more than 10 fields. I added a test in `proptest-derive/tests/misc.rs` that illustrates the problem. The struct `MoreThanTenFields` fails to compile with master today with the following error:

```
error[E0277]: the trait bound `(_IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any): _IMPL_ARBITRARY_FOR_Foo::_proptest::strategy::Strategy` is not satisfied
  --> proptest-derive/tests/misc.rs:70:24
   |
70 | #[derive(Clone, Debug, Arbitrary)]
   |                        ^^^^^^^^^ the trait `_IMPL_ARBITRARY_FOR_Foo::_proptest::strategy::Strategy` is not implemented for `(_IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any)`
   |
   = note: required because of the requirements on the impl of `_IMPL_ARBITRARY_FOR_Foo::_proptest::strategy::Strategy` for `_IMPL_ARBITRARY_FOR_Foo::_proptest::strategy::Map<(_IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any, _IMPL_ARBITRARY_FOR_Foo::_proptest::bool::Any), fn((bool, bool, bool, bool, bool, bool, bool, bool, bool, bool, bool)) -> MoreThanTenFields>`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `proptest-derive`.
```
The error is due to the fact that struct fields are represented as a tuple and `proptest/src/arbitrary/tuples.rs` implements Arbitrary for tuples up to 10 elements.

This change represent structs with more than 10 fields as nested tuples of up to 10 fields, so that the code generated for the test ends up being:

```
cargo rustc --test misc --  -Z unstable-options --pretty=expanded

#[allow(non_upper_case_globals)]
const _IMPL_ARBITRARY_FOR_MoreThanTenFields: () =
    {
        extern crate proptest as _proptest;
        impl _proptest::arbitrary::Arbitrary for MoreThanTenFields {
            type
            Parameters
            =
            (<bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             <bool as _proptest::arbitrary::Arbitrary>::Parameters,
             (<bool as _proptest::arbitrary::Arbitrary>::Parameters,
              <bool as _proptest::arbitrary::Arbitrary>::Parameters));
            type
            Strategy
            =
            _proptest::strategy::Map<((<bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       <bool as
                                       _proptest::arbitrary::Arbitrary>::Strategy,
                                       (<bool as
                                        _proptest::arbitrary::Arbitrary>::Strategy,
                                        <bool as
                                        _proptest::arbitrary::Arbitrary>::Strategy))),
                                     fn((bool, bool, bool, bool, bool, bool,
                                         bool, bool, bool, (bool, bool)))
                                         -> Self>;
            fn arbitrary_with(_top: Self::Parameters) -> Self::Strategy {
                {
                    let (param_0, param_1, param_2, param_3, param_4, param_5,
                         param_6, param_7, param_8, (param_9, param_10)) =
                        _top;
                    _proptest::strategy::Strategy::prop_map((_proptest::arbitrary::any_with::<bool>(param_0),
                                                             _proptest::arbitrary::any_with::<bool>(param_1),
                                                             _proptest::arbitrary::any_with::<bool>(param_2),
                                                             _proptest::arbitrary::any_with::<bool>(param_3),
                                                             _proptest::arbitrary::any_with::<bool>(param_4),
                                                             _proptest::arbitrary::any_with::<bool>(param_5),
                                                             _proptest::arbitrary::any_with::<bool>(param_6),
                                                             _proptest::arbitrary::any_with::<bool>(param_7),
                                                             _proptest::arbitrary::any_with::<bool>(param_8),
                                                             (_proptest::arbitrary::any_with::<bool>(param_9),
                                                              _proptest::arbitrary::any_with::<bool>(param_10))),
                                                            |(tmp_0, tmp_1,
                                                              tmp_2, tmp_3,
                                                              tmp_4, tmp_5,
                                                              tmp_6, tmp_7,
                                                              tmp_8,
                                                              (tmp_9,
                                                               tmp_10))|
                                                                MoreThanTenFields{a:
                                                                                      tmp_0,
                                                                                  b:
                                                                                      tmp_1,
                                                                                  c:
                                                                                      tmp_2,
                                                                                  d:
                                                                                      tmp_3,
                                                                                  e:
                                                                                      tmp_4,
                                                                                  f:
                                                                                      tmp_5,
                                                                                  g:
                                                                                      tmp_6,
                                                                                  h:
                                                                                      tmp_7,
                                                                                  i:
                                                                                      tmp_8,
                                                                                  j:
                                                                                      tmp_9,
                                                                                  k:
                                                                                      tmp_10,})
                }
            }
        }
    };
```

I feel like I should conclude this saying that while the fix seems straightforward, this is something I came up with 10 minutes into trying to use proptest for the first time, so apologies if I'm missing something :)

### DIFF
--- a/proptest-derive/src/ast.rs
+++ b/proptest-derive/src/ast.rs
@@ -31,6 +31,11 @@ use crate::util::self_ty;
 /// but for optimality this should follow what `proptest` supports.
 const UNION_CHUNK_SIZE: usize = 9;
 
+/// The `MAX - 1` tuple length `Arbitrary` is implemented for. After this number,
+/// tuples are expanded as nested tuples of up to `MAX` elements. The value should
+/// be kept in sync with the largest impl in `proptest/src/arbitrary/tuples.rs`.
+const NESTED_TUPLE_CHUNK_SIZE: usize = 9;
+
 /// The name of the top parameter variable name given in `arbitrary_with`.
 /// Changing this is not a breaking change because a user is expected not
 /// to rely on this (and the user shouldn't be able to..).
@@ -278,7 +283,7 @@ impl AddAssign<syn::Type> for Params {
 
 impl ToTokens for Params {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        Tuple2(self.0.as_slice()).to_tokens(tokens)
+        NestedTuple(self.0.as_slice()).to_tokens(tokens)
     }
 }
 
@@ -359,11 +364,12 @@ impl ToTokens for Strategy {
             ),
             Value(ty) => quote_append!(tokens, fn() -> #ty ),
             Map(strats) => {
-                let field_tys = self.types();
-                let strats = strats.iter();
+                let types = self.types();
+                let field_tys = NestedTuple(&types);
+                let strats = NestedTuple(&strats);
                 quote_append!(tokens,
-                    _proptest::strategy::Map< ( #(#strats,)* ),
-                        fn( ( #(#field_tys,)* ) ) -> Self
+                    _proptest::strategy::Map< ( #strats ),
+                        fn( #field_tys ) -> Self
                     >
                 )
             }
@@ -456,7 +462,10 @@ impl ToTokens for ToReg {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         match *self {
             ToReg::Range(to) if to == 1 => param(0).to_tokens(tokens),
-            ToReg::Range(to) => Tuple((0..to).map(param)).to_tokens(tokens),
+            ToReg::Range(to) => {
+                let params: Vec<_> = (0..to).map(param).collect();
+                NestedTuple(&params).to_tokens(tokens)
+            },
             ToReg::API => call_site_ident(API_PARAM_NAME).to_tokens(tokens),
         }
     }
@@ -498,18 +507,55 @@ impl ToTokens for Ctor {
                     _proptest::strategy::LazyJust::new(move || #expr)
                 )
             ),
-            Map(ctors, closure) => {
-                let ctors = ctors.iter();
-                quote_append!(tokens,
-                    _proptest::strategy::Strategy::prop_map(
-                        ( #(#ctors,)* ),
-                        #closure
-                    )
-                );
-            }
+            Map(ctors, closure) => map_ctor_to_tokens(tokens, &ctors, closure),
             Union(ctors) => union_ctor_to_tokens(tokens, ctors),
         }
     }
+}
+
+struct NestedTuple<'a, T>(&'a [T]);
+
+impl<'a, T: ToTokens> ToTokens for NestedTuple<'a, T> {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let NestedTuple(elems) = self;
+        if elems.is_empty() {
+            quote_append!(tokens, ());
+        }
+        else if let [x] = elems {
+            x.to_tokens(tokens);
+        } else {
+            let chunks = elems.chunks(NESTED_TUPLE_CHUNK_SIZE);
+            Recurse(&chunks).to_tokens(tokens);
+        }
+
+        struct Recurse<'a, T: ToTokens>(&'a ::std::slice::Chunks<'a, T>);
+
+        impl<'a, T: ToTokens> ToTokens for Recurse<'a, T> {
+            fn to_tokens(&self, tokens: &mut TokenStream) {
+                let mut chunks = self.0.clone();
+                if let Some(head) = chunks.next() {
+                    if let [c] = head {
+                        // Only one element left - no need to nest.
+                        quote_append!(tokens, #c);
+                    } else {
+                        let tail = Recurse(&chunks);
+                        quote_append!(tokens, (#(#head,)* #tail));
+                    }
+                }
+            }
+        }
+    }
+}
+
+fn map_ctor_to_tokens(tokens: &mut TokenStream, ctors: &[Ctor], closure: &MapClosure) {
+    let ctors = NestedTuple(ctors);
+
+    quote_append!(tokens,
+        _proptest::strategy::Strategy::prop_map(
+            #ctors,
+            #closure
+        )
+    );
 }
 
 /// Tokenizes a weighted list of `Ctor`.
@@ -692,7 +738,7 @@ impl ToTokens for MapClosure {
 
         let MapClosure(path, fields) = self;
         let count = fields.len();
-        let tmps = (0..count).map(tmp_var);
+        let tmps: Vec<_> = (0..count).map(tmp_var).collect();
         let inits = fields.iter().enumerate().map(|(idx, field)| {
             let tv = tmp_var(idx);
             if let Some(name) = &field.ident {
@@ -702,7 +748,8 @@ impl ToTokens for MapClosure {
                 quote_spanned!(field.span()=> #name: #tv )
             }
         });
-        quote_append!(tokens, |( #(#tmps,)* )| #path { #(#inits),* } );
+        let tmps = NestedTuple(&tmps);
+        quote_append!(tokens, | #tmps | #path { #(#inits),* } );
     }
 }
 
--- a/proptest-derive/tests/misc.rs
+++ b/proptest-derive/tests/misc.rs
@@ -67,6 +67,7 @@ enum Quux {
     },
 }
 
+
 #[test]
 fn asserting_arbitrary() {
     fn assert_arbitrary<T: Arbitrary>() {}
--- a/proptest-derive/tests/struct.rs
+++ b/proptest-derive/tests/struct.rs
@@ -1,0 +1,120 @@
+// Copyright 2019 The proptest developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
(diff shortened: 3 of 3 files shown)


proptest-derive/tests/misc.rs
```diff
@@ -67,6 +67,21 @@ enum Quux {
     },
 }
 
+#[derive(Clone, Debug, Arbitrary)]
+struct MoreThanTenFields {
```
Centril: Can you move this to a new file `structs.rs` and test some more edge cases (see the enums.rs test for the general structure for the tests)?

proptest-derive/src/ast.rs
```diff
@@ -30,6 +30,7 @@ use crate::util::self_ty;
 /// Keeping this lower than what `proptest` supports will also work
 /// but for optimality this should follow what `proptest` supports.
 const UNION_CHUNK_SIZE: usize = 9;
+const NESTED_TUPLE_CHUNK_SIZE: usize = 9;
```
Centril: This could also use a comment like the one above (but briefer).

proptest-derive/src/ast.rs
```diff
@@ -278,7 +279,7 @@ impl AddAssign<syn::Type> for Params {
 
 impl ToTokens for Params {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        Tuple2(self.0.as_slice()).to_tokens(tokens)
+        NestedTuple(self.0.as_slice()).to_tokens(tokens)
```
Centril: Why change the expansion here?

(Seems like this would change the expansion of the `Parameters` associated type for cases like `enum`s as well which is a breaking change.)

proptest-derive/src/ast.rs
```diff
@@ -456,7 +458,10 @@ impl ToTokens for ToReg {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         match *self {
             ToReg::Range(to) if to == 1 => param(0).to_tokens(tokens),
-            ToReg::Range(to) => Tuple((0..to).map(param)).to_tokens(tokens),
+            ToReg::Range(to) => {
+                let params: Vec<_> = (0..to).map(param).collect();
+                NestedTuple(&params).to_tokens(tokens)
```
Centril: Same here?

proptest-derive/src/ast.rs
```diff
@@ -498,20 +503,56 @@ impl ToTokens for Ctor {
...
             Union(ctors) => union_ctor_to_tokens(tokens, ctors),
         }
     }
 }
 
+struct NestedTuple<'a, T>(&'a [T]);
+
+impl<'a, T: ToTokens> ToTokens for NestedTuple<'a, T> {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        if self.0.is_empty() {
```
Centril: For readability, can you unpack with `let NestedTuple(elems) = self;` ?

proptest-derive/src/ast.rs
alessandrod: Because of the bound on `Default`on `Parameters`. `Default` is only implemented for tuples up to 12 elements currently.

How is it a breaking change for enums?

alessandrod: (Thanks for reviewing this so quickly btw! :))

Centril: > Because of the bound on `Default`on `Parameters`. `Default` is only implemented for tuples up to 12 elements currently.

Oh... that... I see.

In hindsight it might have been a mistake to auto-generate the `type Parameters` and pass values of it around without the user having asked for it specifically. It adds enormous complexity to the deriving logic and has lead to bugs. Changing the expansion for > 12 also makes it hard to have an intuition about the type `Parameters` is at all.

I would probably be best to just default to `type Parameters = ();` and then change it if the user explicitly requests so, possibly with a `#[proptest(with = "$param_expr")]` (https://github.com/AltSysrq/proptest/pull/104#discussion_r230581497). This would also do away with `no_params`.

For now we should probably keep your change as-is and then consider a more holistic breaking change with a redesign of `type Parameters` in `proptest_derive`.

cc @sunshowers and @AltSysrq if you have opinions :)

> How is it a breaking change for enums?

I re-read the code and no, it cannot be a breaking change. (My bad!) If you would generate too many parameters for an `enum` you'd run into exactly the same error as for tuples (size > 12).

> (Thanks for reviewing this so quickly btw! :))

❤️

proptest-derive/tests/misc.rs
alessandrod: I've done this and I've added the T13 case for the `Parameters` type. Are there any other edge cases you have in mind? (Empty structs are already checked for in `units.rs`)

Centril: It's probably fine like this but if you want to be more thorough it might be good to test that you've exercised `if let [c] = head {` as well as two recursions down, e.g. `9 + 9 + 2` and also `9 + 9 + 1`.

alessandrod: Good point! I've also added a test for a flat, non nested expansion.

proptest-derive/src/ast.rs
sunshowers: Yeah, I would strongly prefer the default to be `no_params` as well.

alessandrod: > For now we should probably keep your change as-is and then consider a more holistic breaking change with a redesign of `type Parameters` in `proptest_derive`.

Thoughts on doing this? I'm trying to publish https://github.com/alessandrod/nuts to crates.io and thought I could depend on my own git fork but apparently crates.io doesn't allow that.

Eh2406: BTW, It is a dev dependency, you could just comment it out when you publish. approximately Crater is the only user of the dev dependencies section of a published crate.

IRog:
It'd be great if we could get this merged in.

PR #155, opened 2019-07-02
Prevent proptest bodies with implicit return values

### LINKED ISSUE #154: Please consider disallow returning values from a proptest
This will happily spin for a while and never fail:

```rust
use proptest::prelude::*;

proptest! {
    #[test]
    fn test_dummy(a: i32) {
        false
    }
}
```

This is the reduced version of my tests which I ported from quickcheck. I’ve been slowly introducing bugs because I forgot to change the returned boolean into an assert. This is something that a “normal” Rust function would have prevented because my function return type doesn’t match the returned value.

shepmaster:
Fixes #154

### DIFF
--- a/proptest/src/sugar.rs
+++ b/proptest/src/sugar.rs
@@ -955,7 +955,7 @@ macro_rules! proptest_helper {
             $($mod)* |$crate::sugar::NamedArguments(
                 _, $crate::proptest_helper!(@_WRAPPAT ($($parm),*)))|
             {
-                $body;
+                let _: () = $body;
                 Ok(())
             })
         {
@@ -975,7 +975,7 @@ macro_rules! proptest_helper {
             $($mod)* |$crate::sugar::NamedArguments(
                 _, $crate::proptest_helper!(@_EXT _PAT ($($arg)*)))|
             {
-                $body;
+                let _: () = $body;
                 Ok(())
             })
         {


shepmaster:
Feel free to close. I was curious as to why this constitutes a breaking change so I wanted to see the failing tests, but none of them failed; perhaps an indication of more tests that need to be written?

shepmaster:
This does appear to prevent my case from occurring:

```diff
diff --git a/proptest/src/sugar.rs b/proptest/src/sugar.rs
index 8f887ef..42c9c8b 100644
--- a/proptest/src/sugar.rs
+++ b/proptest/src/sugar.rs
@@ -1635,3 +1635,13 @@ mod any_tests {
         let proptest_helper!(@_EXT _PAT( _x in 1..2 )) = 1;
     }
 }
+
+#[cfg(test)]
+mod compile_fail_goes_somewhere {
+    proptest! {
+        #[test]
+        fn disallows_returning_boolean(_: i32) {
+            true
+        }
+    }
+}
```

```
error[E0308]: mismatched types
    --> proptest/src/sugar.rs:1644:13
     |
1644 |             true
     |             ^^^^ expected (), found bool
     |
     = note: expected type `()`
                found type `bool`
```

AltSysrq:
The breakage is very much an edge case and is something I think _should_ break, but is nonetheless breaking. For example, adapted from the dateparser example:

```rust
fn parse_date(s: &str) -> Option<(u32, u32, u32)> { /* code */ }
proptest! {
  #[test] 
  fn doesnt_crash(s in "\\PC*") {
    parse_date(&s) // Oops, forgot semicolon
  }
}
```

In this case, the accidental return value that gets discarded was actually intended to be discarded so it doesn't affect the validity of the test.

shepmaster:
As I said, feel free to close this.

As a user, I think being forced to add a semicolon here is not truly a breaking change as nothing was “supposed” to be returned in the first place. I’d make the argument that it’s a bug in the macro.

PR #156, opened 2019-07-08
Added link to book in README

jean-airoldie:
The book is great for newcomers, why not make it more visible?


matthew-russo:
This is already in the README:
https://github.com/proptest-rs/proptest/blob/master/proptest/README.md?plain=1#L9

where it shows up as a badge with the title "website"

PR #165, opened 2019-09-03
Add missing ?Sized bound to Arbitrary for Cow

brunocodutra:


proptest/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
```
Centril: Seems more like a minor feature improvement than a bug-fix?

brunocodutra: I reckoned the intention has always been to allow for references of unsized types, as there's no obvious reason why not, but I'm fine with relabeling this change.

Issue #153, opened 2019-07-01
#[proptest] Attribute Macro


davidbarsky:
Hi! I wrote a _kinda_ rough implemenation of a `#[proptest]` attribute macro in https://github.com/awslabs/aws-lambda-rust-runtime/pull/111. It accepts an `impl Strategy<Value = T>`, and I wanted to see if there's interest in me sending a pull request to introduce it here. It looks a bit like this:

```rust
pub fn gen_function_arn() -> impl Strategy<Value = FunctionArn> {
    let expr = "arn:aws:lambda:us-east-1:[0-9]{12}:function:custom-runtime";
    let arn = string_regex(expr).unwrap();
    arn.prop_map(FunctionArn)
}

#[proptest(gen_function_arn())]
fn function_arn(arn: FunctionArn) {
    let mut headers = HeaderMap::new();
    headers.typed_insert(arn.clone());
    prop_assert_eq!(headers.typed_get::<FunctionArn>(), Some(arn));
}
```

Centril:
cc https://github.com/AltSysrq/proptest/issues/149 and https://github.com/AltSysrq/proptest/issues/97#issuecomment-433246156.

A `#[proptest]` attribute would be neat I think.

---------------------------

We'll need to think about the details like:
- how to most ergonomically provide custom strategies
- how configurations should be passed
- how function return types should be dealt with in terms of `Termination` or some other trait
- whether closure style invocation can be supported, 

Also, if you just write:

```rust
#[proptest]
fn addition_is_commutative(x: u8, y: u8) {
   ... 
}
```
then presumably `any::<u8>()` will be used for both.

---------------------------

```rust
#[proptest(gen_function_arn())]
```

The notation here for passing custom strategies seems suboptimal.

It could be improved with https://github.com/rust-lang/rust/issues/60406 but that isn't stable yet (tho one some bugs are fixed it could be stabilized relatively quickly after baking for 6 weeks). This would allow e.g.:

```rust
#[proptest]
fn function_arn(
    #[strategy = gen_function_arn()]
    arn: FunctionArn
) {
```

We also cannot support the `fn foo(pat in expr) {` notation either with `#[proptest]` as parsing is done before attributes are expanded and the type must always be provided lest we allow `pat ::= ... | pat ":" type ;` into Rust's pattern grammar (which could possibly be done as part of https://github.com/rust-lang/rfcs/pull/2522).

davidbarsky:
> We'll need to think about the details like:
[...]

PR #166, opened 2019-09-03
WIP implementation of proptest-attributes

davidbarsky:
(cc: @Centril) 

A WIP implementation of the the proc macros discussed in https://github.com/AltSysrq/proptest/issues/153. This currently supports:

- A single argument + strategy per test.

This does not support:

- Everything else discussed in https://github.com/AltSysrq/proptest/issues/153.

That said, _some_ feedback is better than none, so here we are. The changes are better seen than described, so take a look at the modified `parse_date` example, located in `proptest-attributes/tests/hello.rs`:

```rust
use proptest::{
    prelude::*, prop_assert_eq, strategy::Strategy, string::string_regex,
};
use proptest_attributes::proptest;

fn parse_date(s: &str) -> Option<(u32, u32, u32)> {
    if 10 != s.len() {
        return None;
    }

    // NEW: Ignore non-ASCII strings so we don't need to deal with Unicode.
    if !s.is_ascii() {
        return None;
    }

    if "-" != &s[4..5] || "-" != &s[7..8] {
        return None;
    }

    let year = &s[0..4];
    let month = &s[5..7];
    let day = &s[8..10];

    year.parse::<u32>().ok().and_then(|y| {
        month
            .parse::<u32>()
            .ok()
            .and_then(|m| day.parse::<u32>().ok().map(|d| (y, m, d)))
    })
}

fn gen_valid_date() -> impl Strategy<Value = String> {
    let expr = "[0-9]{4}-[0-9]{2}-[0-9]{2}";
    string_regex(expr).unwrap()
}

fn gen_all_utf8() -> impl Strategy<Value = String> {
    let expr = "\\PC*";
    string_regex(expr).unwrap()
}

prop_compose! {
  fn gen_parsed_date()(year in 0u32..10000, month in 1u32..13, day in 1u32..32) -> (u32, u32, u32) {
    (year, month, day)
  }
}

#[proptest(gen_valid_date())]
fn parses_all_valid_dates(s: String) {
    parse_date(&s).unwrap();
}

#[proptest(gen_all_utf8())]
fn doesnt_crash(s: String) {
    parse_date(&s);
}

#[proptest(gen_parsed_date())]
fn parses_date_back_to_original(date_tuple: (u32, u32, u32)) {
    let (y, m, d) = date_tuple;
    let (y2, m2, d2) =
        parse_date(&format!("{:04}-{:02}-{:02}", y, m, d)).unwrap();
    // prop_assert_eq! is basically the same as assert_eq!, but doesn't
    // cause a bunch of panic messages to be printed on intermediate
    // test failures. Which one to use is largely a matter of taste.
    prop_assert_eq!((y, m, d), (y2, m2, d2));
}
```

I think that, beyond what was described in #153, there might be an opportunity to simplify the `prop_compose!` macro with a proc macro.

### DIFF
--- a/proptest-attributes/src/lib.rs
+++ b/proptest-attributes/src/lib.rs
@@ -20,21 +20,21 @@ pub fn proptest(
 
     match inputs.len() {
         1 => {
-            let arg = match inputs.first().unwrap() {
-                FnArg::Typed(arg) => arg,
+            let param = match inputs.first().unwrap() {
+                FnArg::Typed(param) => param,
                 FnArg::Receiver(recv) => {
                     let tokens = quote_spanned! { recv.span() =>
-                        compile_error!("The #[proptest] macro cannot be applied to a method.");
+                        compile_error!("The `#[proptest]` macro cannot be applied to a method.");
                     };
                     return TokenStream::from(tokens);
                 }
             };
-            let arg_name = &arg.pat;
+            let param_name = &param.pat;
             quote! {
                 #[test]
                 #(#attrs)*
                 fn #name() #ret {
-                    proptest::proptest!(|(#arg_name in #expr)| {
+                    proptest::proptest!(|(#param_name in #expr)| {
                         #body
                     })
                 }
@@ -42,7 +42,7 @@ pub fn proptest(
         }
         _ => {
             let tokens = quote_spanned! { input.sig.span() =>
-                compile_error!("The #[proptest] macro can only be applied to a function with a single argument.");
+                compile_error!("The `#[proptest]` macro can only be applied to a function with a single argument.");
             };
             return TokenStream::from(tokens);
         }
--- a/proptest-attributes/src/lib.rs
+++ b/proptest-attributes/src/lib.rs
@@ -1,0 +1,51 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use quote::{quote, quote_spanned};
+use syn::{spanned::Spanned, Expr, FnArg, ItemFn};
+
+#[proc_macro_attribute]
+pub fn proptest(
+    attr: TokenStream,
+    item: TokenStream,
+) -> proc_macro::TokenStream {
+    let input = syn::parse_macro_input!(item as ItemFn);
+    let expr = syn::parse_macro_input!(attr as Expr);
+
+    let ret = &input.sig.output;
+    let name = &input.sig.ident;
+    let body = &input.block;
+    let attrs = &input.attrs;
+    let inputs = &input.sig.inputs;
+
+    match inputs.len() {
+        1 => {
+            let arg = match inputs.first().unwrap() {
+                FnArg::Typed(arg) => arg,
+                FnArg::Receiver(recv) => {
+                    let tokens = quote_spanned! { recv.span() =>
+                        compile_error!("The #[proptest] macro cannot be applied to a method.");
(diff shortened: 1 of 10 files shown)


proptest-attributes/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+    let attrs = &input.attrs;
+    let inputs = &input.sig.inputs;
+
+    match inputs.len() {
+        1 => {
+            let arg = match inputs.first().unwrap() {
+                FnArg::Typed(arg) => arg,
+                FnArg::Receiver(recv) => {
+                    let tokens = quote_spanned! { recv.span() =>
+                        compile_error!("The #[proptest] macro cannot be applied to a method.");
```
Centril:
```suggestion
                        compile_error!("The `#[proptest]` macro cannot be applied to a method.");
```

proptest-attributes/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+                fn #name() #ret {
+                    proptest::proptest!(|(#arg_name in #expr)| {
+                        #body
+                    })
+                }
+            }
+        }
+        _ => {
+            let tokens = quote_spanned! { input.sig.span() =>
+                compile_error!("The #[proptest] macro can only be applied to a function with a single argument.");
```
Centril:
```suggestion
                compile_error!("The `#[proptest]` macro can only be applied to a function with a single argument.");
```

proptest-attributes/tests/compile-fail/method.stderr
```diff
@@ -0,0 +1,11 @@
+error: The #[proptest] macro cannot be applied to a method.
+  --> $DIR/method.rs:17:21
+   |
+17 |     fn function_arn(&self) {
+   |                     ^^^^^
+
+error[E0601]: `main` function not found in crate `$CRATE`
```
Centril: off-topic error, add `fn main() {}` to get rid of it. (we do this in the compiler)

proptest-attributes/tests/compile-fail/multiple-args.stderr
```diff
@@ -0,0 +1,11 @@
+error: The #[proptest] macro can only be applied to a function with a single argument.
+  --> $DIR/multiple-args.rs:10:1
+   |
+10 | fn parse_date(date: String, _second_arg: String) {}
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error[E0601]: `main` function not found in crate `$CRATE`
```
Centril: same here, off-topic error.

davidbarsky: Fixed.

proptest-attributes/src/lib.rs
davidbarsky: Fixed, thanks!

davidbarsky:
Hi! @AltSysrq! I think this is ready to review! I'll be happy to change the authorship of the crate to "proptest contributers".

matthew-russo:
This got deleted when I switched from branch master to main. We have added an attribute macro in `proptest-macro` crate. https://crates.io/crates/proptest-macro

If there are specific requests from this historical thread, feel free to cut an issue and we can add it in the new `proptest-macro` crate

PR #175, opened 2019-11-22
Add support for hardware entropy source on x86-64.

gz:
This allows `no_std` builds to use the `rdrand` instruction on x86
to generate a random seed for the random number generators.

An example invocation to enable it be:
```
cargo +nightly run --example fib --features alloc --features unstable --features hardware-rng --no-default-features
```


gz:
Is this something the maintainers would be interested in merging or out of scope for the project (we're using this patch for a project so if I could merge it upstream into proptest I don't have to maintain a fork anymore)?

PR #194, opened 2020-05-14
avoiding panic in passthrough RNG

mimoo:
There is no reason to panic if the seed passed to a `PassThrough` RNG gets depleted. We can just return 0s instead.

### DIFF
--- a/proptest/src/test_runner/rng.rs
+++ b/proptest/src/test_runner/rng.rs
@@ -43,10 +43,7 @@ pub enum RngAlgorithm {
     /// This is useful when Proptest is being driven from some other entropy
     /// source, such as a fuzzer.
     ///
-    /// It is the user's responsibility to ensure that the seed is "big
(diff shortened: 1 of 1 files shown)


AltSysrq:
I have this staged to the `0.10.0-changes` branch. It's not going in today since the appveyor tests take forever, so probably tomorrow evening US time.

I did simplify the logic here a bit: https://github.com/AltSysrq/proptest/commit/475dd442ae1da5b65659c05c8b4bd77c7e5ca0e1

mimoo:
thanks @AltSysrq !

sunshowers:
I'll try and submit a PR shortly to remove references to `TupleUnion` and replace them with `LazyTupleUnion` for the 0.10 branch. Would be worth getting in for API cleanliness.

sunshowers:
@AltSysrq See #195 for the removal of the old TupleUnion. It's a breaking change so I hope it can make it in time for 0.10! :)

PR #195, opened 2020-05-28
Replace TupleUnion with LazyTupleUnion

sunshowers:
LazyTupleUnion was added in #144 to improve performance of union
strategies. This commit removes the old TupleUnion impl and
replaces it with LazyTupleUnion.


AltSysrq:
Thanks! Impeccable timing too, the appveyor build was just about done.

PR #202, opened 2020-07-22
New RngAlgorithm::Recorder to record bytes used to create value

mimoo:
This change is useful for fuzzing,
as we need to know what bytes can be used with a passthrough RNG in order to create a valid corpus.

This way, we can generate a value with the Record RNG and save this as corpus for fuzzing:

```rust
let config = test_runner::Config::default();
let recorder_rng = TestRng::default_rng(RngAlgorithm::Recorder);
let mut runner = TestRunner::new_with_rng(config, recorder_rng);

let strategy = some_strategy();
let strategy_tree = match strategy.new_tree(&mut runner) {
    Ok(x) => x,
    Err(_) => return,
};
let data = strategy_tree.current();
let bytes_used = runner.bytes_used();
```

### DIFF
--- a/proptest/src/test_runner/rng.rs
+++ b/proptest/src/test_runner/rng.rs
@@ -50,6 +50,12 @@ pub enum RngAlgorithm {
     /// entropy is actually needed. This means that combinators like
     /// `prop_perturb` and `prop_flat_map` can require extremely large inputs.
     PassThrough,
+    /// This is equivalent to the ChaCha Rng, with the addition that it records,
+    /// the bytes used to create a value.
+    ///
+    /// This is useful when Proptest is used for fuzzing, and a corpus of initial
+    /// inputs need to be created.
+    Recorder,
     #[allow(missing_docs)]
     #[doc(hidden)]
     _NonExhaustive,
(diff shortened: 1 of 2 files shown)


proptest/src/test_runner/rng.rs
```diff
@@ -330,6 +406,14 @@ impl TestRng {
         TestRng::from_seed_internal(Seed::from_bytes(algorithm, seed))
     }
 
+    /// dumps the bytes obtained from the RNG so far (only works if the RNG is set to Recorder)
+    pub fn bytes_used(&self) -> Vec<u8> {
```
AltSysrq, line 410: Wouldn't it be better to return `Option<Vec<u8>>` here or something?

proptest/src/test_runner/runner.rs
```diff
@@ -371,6 +371,11 @@ impl TestRunner {
         &self.config
     }
 
+    /// dump the bytes used to produce a value
+    pub fn bytes_used(&self) -> std::vec::Vec<u8> {
```
AltSysrq, line 375: This needs to use `crate::std_facade::Vec` for `no_std` compatibility

proptest/src/test_runner/rng.rs
mimoo: I couldn't feel of a use-case where you wouldn't `unwrap()` if you did pass an `Option`. Because in the end you know that you've generated a value before calling this one. Wdyt?

AltSysrq: Yeah, I guess that's reasonable

AltSysrq:
Published in 0.10.1.

PR #207, opened 2020-08-26
Update to version 2 of quick-error

LovecraftianHorror:
This just updates `quick-error` to version 2. This did involve switching `cause` over to `source` and removing `description` since both were deprecated since rust 1.33.0 and 1.42.0 respectively. (Deja vu)

I noticed there are unrelated problems trying to test `proptest-derive` which will probably cause CI to fail.


LovecraftianHorror:
I see now from `CONTRIBUTING.md` that nightly is required for working on `proptest_derive`

PR #210, opened 2020-09-11
Book: Getting-started: Mention `cargo test`

siedentop:
It was not clear to me that proptest has to be run with `cargo test`. Maybe obvious but it did confuse me -- and maybe it confuses others as well?


siedentop:
Closing due to inactivity.

PR #217, opened 2020-10-25
Fix shuffle algorithm

### LINKED ISSUE #216: prop_shuffle() doesn't generate all permutations
The following code prints out each unique permutation generated by the `prop_shuffle()` combinator on an input vec with elements [1, 2, 3]:

```rust
use proptest::{
    prelude::*,
    test_runner::{Config, TestRunner},
};
use std::{collections::HashSet, sync::Mutex};

fn main() {
    let mut runner = TestRunner::new(Config {
        cases: 1_000_000,
        ..Config::default()
    });

    let input = vec![1, 2, 3];
    let seen = Mutex::new(HashSet::new());
    runner
        .run(&(Just(input).prop_shuffle()), |ints| {
            let mut seen = seen.lock().unwrap();
            if !seen.contains(&ints) {
                println!("{:?}", ints);
                seen.insert(ints);
            }

            Ok(())
        })
        .unwrap();
}
```

When ran, this is the output:

```
[3, 1, 2]
[2, 3, 1]
```

Only two permutations are generated, instead of the six I expected.

---

rustc 1.48.0-nightly (8876ffc92 2020-10-02)
proptest v0.10.1

Seeker14491:
It must be possible for no swap to happen at each step in order to be able to generate all permutations.

Fixes #216

### DIFF
--- a/proptest/src/strategy/shuffle.rs
+++ b/proptest/src/strategy/shuffle.rs
@@ -170,7 +170,7 @@ where
             // Determine the other index to be swapped, then skip the swap if
             // it is too far. This ordering is critical, as it ensures that we
             // generate the same sequence of random numbers every time.
-            let end_index = rng.gen_range(start_index + 1, len);
+            let end_index = rng.gen_range(start_index, len);
             if end_index - start_index <= max_swap {
                 value.shuffle_swap(start_index, end_index);
             }


Andlon:
I'm currently running into this issue. Unless there's an obstacle, could we please get this merged?

AltSysrq:
Thanks for the fix, and sorry for being unresponsive. This is in 1.0.0.

PR #234, opened 2021-03-16
Add Config::with_failure_persistence

mzabaluev:
A field record update of the `failure_persistence` member of `Config` is a handful to write in the `proptest_config` attribute and needs all the gory details of the field type initialization.

Add a convenience constructor making use of a generic parameter over `FailurePersistence` impls.

### DIFF
--- a/proptest/src/test_runner/config.rs
+++ b/proptest/src/test_runner/config.rs
@@ -419,6 +419,32 @@ impl Config {
         result
     }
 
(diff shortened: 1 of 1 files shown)


matthew-russo:
This got closed when i switched branch from `master` to `main`. I'll re-open against the main branch. I'd rather switch to just having a config builder that's a bit more generic I think

matthew-russo:
reopened at #508

PR #243, opened 2021-07-03
Use const generics to remove 32 element array size restriction

jvff:
This PR replaces the macro generated code for small arrays (with less than 32 elements) with const-generic code that have no size restrictions. I'm not sure if this is a desired feature, but I found it could be useful to me so I gave a shot at trying to implement it. The `uniform{1..32}` functions are still generated as "proxy functions" in order to keep backwards compatibility.

### DIFF
--- a/proptest/src/arbitrary/arrays.rs
+++ b/proptest/src/arbitrary/arrays.rs
@@ -12,27 +12,12 @@
 use crate::arbitrary::{any_with, Arbitrary};
 use crate::array::UniformArrayStrategy;
 
-macro_rules! array {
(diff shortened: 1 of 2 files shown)


sunshowers:
Were you interested in taking this PR @AltSysrq? I'd love to be able to use const generics for something I'm building.

matthew-russo:

matthew-russo:
This is already addressed via:
https://github.com/proptest-rs/proptest/commit/b47d42bf78e72412ace7dfc435adaf7a1506a72b

PR #255, opened 2021-11-09
Allow trailing comma in prop_assert_eq/ne like std

louy2:


matthew-russo:

PR #261, opened 2022-02-24
Alternative uniform float samplers

### LINKED ISSUE #258: Test failures since switch to rand 0.8
I am investigating upgrading the rust packaging in Debian to rand 0.8 and I have noticed that proptest seems to suffer from test failures since it switched to rand 0.8.

cargo test -p proptest --lib

failures:

---- num::test::contract_sanity::f32::range_from stdout ----
thread 'num::test::contract_sanity::f32::range_from' panicked at 'Uniform::new_inclusive: range overflow', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.8.4/src/distributions/uniform.rs:996:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- num::test::contract_sanity::f32::range_to_inclusive stdout ----
thread 'num::test::contract_sanity::f32::range_to_inclusive' panicked at 'Uniform::new_inclusive: range overflow', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.8.4/src/distributions/uniform.rs:996:1

---- num::test::contract_sanity::f64::range_from stdout ----
thread 'num::test::contract_sanity::f64::range_from' panicked at 'Uniform::new_inclusive: range overflow', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.8.4/src/distributions/uniform.rs:997:1

---- num::test::contract_sanity::f64::range_to_inclusive stdout ----
thread 'num::test::contract_sanity::f64::range_to_inclusive' panicked at 'Uniform::new_inclusive: range overflow', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.8.4/src/distributions/uniform.rs:997:1

failures:
    num::test::contract_sanity::f32::range_from
    num::test::contract_sanity::f32::range_to_inclusive
    num::test::contract_sanity::f64::range_from
    num::test::contract_sanity::f64::range_to_inclusive

test result: FAILED. 1421 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 783.81s

error: test failed, to rerun pass '-p proptest --lib'

I can reproduce the failures with commit commit  bc2d867a7cd06060e0b565b72fce1d94cc2fe943 (current HEAD) and commit 291908ac22a4c05f9fcad31cb460c502723b22ae (the first commit to use rand 0.8) and but not with 37dba3abe9aa0a1cc6b9762353bf6dd89b9ef6a5 (the last commit to use rand 0.7)

Ideally I'd like to see a fix for this issue, but if that is not possible, i'd at least like an opinion on how serious these failures are so I can decide whether they should block Debian's transition to rand 0.8.

mikerite:
Since 0.8.4 the uniform float samplers provided by the rand crate have
panicked when `low - high` overflows. We fix this problem by
implementing our own samplers.

Fixes #258


proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,318 @@
...
+
+                fn new<B1, B2>(low: B1, high: B2) -> Self
+                where
+                    B1: SampleBorrow<Self::X> + Sized,
+                    B2: SampleBorrow<Self::X> + Sized,
+                {
+                    let low = low.borrow().0;
+                    let high = high.borrow().0;
+
+                    let values = SampleValueCollection::new(low, last_float(high));
```
matthew-russo, line 77: I'm pretty sure theres a minor bug here since we pass in `last_float(high)` here, and then `SampleValueCollection::new` also calculates `last_float(high)` with whats passed in so we actually end up using `last_float(last_float(high))`. Proposing we remove `SampleValueCollection::new`, only have `SampleValueCollection::new_inclusive`, and keep passing in `last_float(high)`here

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,318 @@
...
+            const MAX_PRECISE_INT: $int_typ =
+                (2 as $int_typ).pow($typ::MANTISSA_DIGITS);
+
+            // The collection of sample values that may be generated by UniformF32U.
+            impl SampleValueCollection {
+                fn new(low: $typ, high: $typ) -> Self {
+                    let high_incl = last_float(high);
+
+                    Self::new_inclusive(low, high_incl)
+                }
```
matthew-russo, line 128: with above comment: propose removing this function and only have `new_inclusive` to prevent possibilities of calculating `last_float(high)` twice.

matthew-russo (changes requested):
Sorry for the delay in review. One small suggestion regarding what I think is a bug -- let me know if I'm misunderstanding anything

matthew-russo:
Can we also update the CHANGELOG.md (https://github.com/proptest-rs/proptest/blob/master/CONTRIBUTING.md#updating-the-changelog) accordingly? We'll need a new `## Unreleased` section.

cameron1024 (approved):
Looks good, thanks for the PR 🎉

matthew-russo (approved):
Looks good to me, thanks for the quick update!

matthew-russo:
@russelldb can you give this a look when you get a chance?

russelldb (approved):
Looks good to me, though as I said, new here, and not au fait yet with the macro heavy style of proptest. I raised a question about API. FWIW, I merged this is my fork a long time ago and have used it for some time, so I would say it appears to work in practice.

proptest/src/num.rs
```diff
@@ -12,16 +12,18 @@
...
+
 use crate::test_runner::TestRunner;
-use core::ops::Range;
 use rand::distributions::uniform::{SampleUniform, Uniform};
 use rand::distributions::{Distribution, Standard};
 
 pub(crate) fn sample_uniform<X: SampleUniform>(
     run: &mut TestRunner,
-    range: Range<X>,
+    start: X,
```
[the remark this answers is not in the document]
matthew-russo: This is a crate-only function so there shouldn't be any impact in users code. I can't think of any functional reasons to change it or any functional changes that come out of it but it does make it consistent with `sample_uniform_incl`.

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+        pub mod $typ {
+            use rand::prelude::*;
+            use rand::distributions::uniform::{
+                SampleBorrow, SampleUniform, Uniform, UniformSampler,
+            };
+
+            #[must_use]
+            // Returns the previous float value. In other words the greatest value representable
+            // as a float such that `last_float(a) < a`. `-0.` is treated as `0.`.
+            fn last_float(a: $typ) -> $typ {
```
rex-remind101, line 28: not sure about naming here, sounds more like `max_float_less_than_or_equal_to(a)` or we could copy swift's convention of `clamping` and call this `float_clamping(a)` or `clamping_float(a)` -- i think either of those work.

sauce https://developer.apple.com/documentation/swift/binaryinteger/init(clamping:)

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+                debug_assert!(a.is_finite() && a > $typ::MIN, "`last_float` invalid input: {}", a);
+                if a == (0.) {
+                    -$typ::from_bits(1)
+                } else if a < 0. {
+                    $typ::from_bits(a.to_bits() + 1)
+                } else {
+                    $typ::from_bits(a.to_bits() - 1)
```
rex-remind101, line 35: It's unclear to me what the -1 and +1 do here, would you mind explaining out of curiosity?

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+                }
+            }
+
+            #[must_use]
+            fn last_gap(a: $typ) -> $typ {
```
rex-remind101, line 40: i suggest `clamped_float_remainder` or something along those lines.

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+                    assert!(index < self.count, "index out of bounds");
+
+                    if index == self.count - 1 {
+                        return self.end;
+                    }
+
+                    ((index / 2) as $typ).mul_add(
+                        2. * self.step,
+                        (index % 2) as $typ * self.step + self.start,
+                    )
```
rex-remind101, line 172: I think this little section of code would benefit from a comment, very unclear to me what it's intent is. Not sure why we'd divide by 2 only to then multiply but 2. I understand that floats are finicky, so maybe you're exploiting something that should be added to the comment.

proptest/src/num/float_samplers.rs
rex-remind101: Ok, after reading through all the code more thoroughly i believe i understand the intent now. This is getting a float value that is the smallest bit increment less in magnitude from the input.

maybe `magnitude_decreased_by_bit(float)` or `minimally_smaller_float(float)`. i can't think of conventional language for this, but we should at least strive for higher clarity. given this is unconventional, i think i'm more in favor of the more explicit `magnitude_decreased_by_bit`.

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+// except according to those terms.
+
+//! Alternative uniform float samplers because the ones provided by the rand crate are prone
+//! to overflow. The samplers work by uniformly selecting from a set of equally spaced values in
+//! the interval and the included bounds. Selection is slighly biased towards the bounds.
+
+pub(crate) use self::f32::F32U;
+pub(crate) use self::f64::F64U;
+
+macro_rules! float_sampler {
```
rex-remind101, line 17: Because of the subtle nuances in this code, i highly suggest a comment about the overall strategy with an example of what's expected in terms of sampling given some range.

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
+//-
+// Copyright 2022 Michael Wright
```
rex-remind101, line 2: The copyright should reflect the org, i.e. proptest-rs. Also, not sure why we have 2 licenses, MIT i think would be perfect on its own, but that's a larger discussion.

proptest/src/num/float_samplers.rs
rex-remind101: given my other comment update, maybe `magnitude_remainder` as a name?

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Alternative uniform float samplers because the ones provided by the rand crate are prone
+//! to overflow. The samplers work by uniformly selecting from a set of equally spaced values in
+//! the interval and the included bounds. Selection is slighly biased towards the bounds.
```
rex-remind101, line 12: oh, my eyes missed this. strategy is covered, but i think an example will help as well. especially given the nuance of `Selection is slighly biased towards the bounds.` which i'm not entirely sure what that means, i would think a `Uniform` sampling would not have bias, it would be uniform across the bounds.

also `slighly` -> `slightly`

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+
+                fn new<B1, B2>(low: B1, high: B2) -> Self
+                where
+                    B1: SampleBorrow<Self::X> + Sized,
+                    B2: SampleBorrow<Self::X> + Sized,
+                {
+                    let low = low.borrow().0;
+                    let high = high.borrow().0;
+
+                    let values = SampleValueCollection::new_inclusive(low, last_float(high));
```
rex-remind101, line 77: if someone were to provide a range between low and high, wouldn't the expectation be that that range sets the bounds for the sampling? by funneling high through `last_float`, it's decreasing the high bounds by 1 bit of magnitude. I'm sure there's a reason for this that i'm not understanding, could you help me better understand? Possible question here is, what does "new_inclusive" imply that "new" does not?

proptest/src/num/float_samplers.rs
matthew-russo: technically our `CONTRIBUTING.md` say to use `either your name or "The proptest developers", at your choice.` We should update `CONTRIBUTING.md` if we only want "The proptest developers"

https://github.com/proptest-rs/proptest/blob/master/CONTRIBUTING.md#copyright-headers

proptest/src/num/float_samplers.rs
```diff
@@ -0,0 +1,338 @@
...
+                    debug_assert!(
+                        max_gaps.floor() == max_gaps,
+                        "max_gaps is an integer"
+                    );
+
+                    let count = if low.signum() == high.signum() {
+                        max_gaps as $int_typ - min_gaps.floor() as $int_typ
+                    } else {
+                        max_gaps as $int_typ + min_gaps.ceil() as $int_typ
+                    } + 1;
```
rex-remind101, line 151: i feel like there's an obvious reason for this, but i can't seem to piece together why we add `1` at the end here, mind explaining?

rex-remind101 (changes requested):
biggest concern is copyright and naming. the rest are asks for clarity.

we're also still failing in ci with `unknown feature `llvm_asm``, was this an existing issue and we should address in a separate pr or is this new?

proptest/src/num/float_samplers.rs
rex-remind101: It seems simpler to just request `The proptest developers` then across the board. People will get the karma through git and github's existing features regardless.

I'll open a pr to update the readme 👍

proptest/src/num/float_samplers.rs
matthew-russo: i think i'd actually prefer `minimally_smaller_float` since it conveys the intent/purpose that we are getting the immediate float that is smaller than this one, rather than the implementation -- that we're doing this by decrementing the bit value by 1. 

that said i dont have overly strong opinions.

proptest/src/num/float_samplers.rs
matthew-russo: I do like the term `gap` though perhaps `difference` is more precise? I'm not a huge fan of `magnitude_remainder` on its own.

following on my preferences from other comment, i'd say something like `difference_between_minimally_smaller_float` or `difference_to_minimally_smaller_float`?

proptest/src/num/float_samplers.rs
matthew-russo: this trait is from Rand so we don't have much semantic control over. That said, UniformSampler provides 2 constructors to account for inclusive and exclusive upper bounds.

`new` samples in a range with exclusive upper bound `[low, high)` or `low..high` in rust range syntax

`new_inclusive` samples in a range with inclusive upper bound `[low, high]` or `low..=high` in rust range syntax

https://docs.rs/rand/latest/rand/distributions/uniform/trait.UniformSampler.html#tymethod.new

matthew-russo:
> was this an existing issue and we should address in a separate pr or is this new?

This is new afaik. I was planning on working on a fix once we get this merged

proptest/src/num/float_samplers.rs
matthew-russo: my comment doesn't make a whole lot of sense since you're specifically referring to `SampleValueCollection`, not the `UniformSampler` methods. 

I like that our internal impls follow the same naming convention which is why I brought up UniformSampler.

Specifically for why we only have `new_inclusive` -- the previous revision of this PR had both `new` and `new_inclusive` but there was a slight bug where we ended up doubling up on calls to `last_float` so i suggested condensing to just a single `new_inclusive` that was unambiguous and just implies the range includes the value passed in as the second parameter

proptest/src/num/float_samplers.rs
rex-remind101: https://github.com/proptest-rs/proptest/pull/280

proptest/src/num/float_samplers.rs
rex-remind101: that's fair, i'll support that 👍

proptest/src/num/float_samplers.rs
rex-remind101: `difference_to_minimally_smaller_float` sounds good to me 👍

proptest/src/num/float_samplers.rs
rex-remind101: ah, thank you that explains a lot

proptest/src/num/float_samplers.rs
matthew-russo: Interesting! I'm in favor of using that -- wasn't aware there was an agreed upon name for this already. Thanks for sharing

rex-remind101: Wasn't aware of that either, 👍

proptest/src/num/float_samplers.rs
rex-remind101: This is clear now, thank you

proptest/src/num/float_samplers.rs
rex-remind101: right, thank you

proptest/src/num/float_samplers.rs
rex-remind101: appears to be common lingo https://www.google.com/search?client=firefox-b-1-d&q=next_down

there is a significant difference here though in that our impl does `next_up` for negatives, but the name at least is much closer to convention.

rex-remind101 (approved):
not sure what timezones ya'll are in. will approve since we found some good naming conventions to update to so you can go ahead with final changes and merging without being blocked 👍 thanks for helping me understand some of the finer nuances

PR #262, opened 2022-03-05
note that PROPTEST_CASES env var requires the std feature

RalfJung:
I just lost a whole bunch of time trying to figure out why setting `PROPTEST_CASES` has no effect, until I read its code and saw that this relies on the `std` feature. Would be better to document that more clearly.


cameron1024 (approved):
LGTM - thanks for spending the time finding this 😁

matthew-russo (approved):

PR #264, opened 2022-04-06
proptest-derive: upgrade to stable proc macro ecosystem

benesch:
Upgrade proptest-derive to the stable versions of proc-macro2, syn, and
quote.


rex-remind101:
Our Cargo.toml is using the stable proc marco ecosystem. Is there anything else needed here?

PR #265, opened 2022-04-07
use u64 instead of u32 for max_shrink_iters

### LINKED ISSUE #254: PROPTEST_MAX_SHRINK_ITERS as u32 is too small
Hey folks, let met just say that first, I love proptest.

One thing is a bit annoying though: I have a few tests with a lots of parameters, and sometimes I cannot completely shrink the examples in u32::MAX iters I need more. And I can't find a way to say "no-limit" to the number of iters, as 0 means "no shrinking".

Is there a workaround for this ?

kali:
Use u64 instead of u32 for max_shrink_iters. I have high cardinality problems (tensors...) and hitting this all the time. I can't see any realistic drawback.

Fixes #254

### DIFF
--- a/proptest/src/strategy/flatten.rs
+++ b/proptest/src/strategy/flatten.rs
@@ -261,7 +261,7 @@ impl<S: Strategy, R: Strategy, F: Fn(S::Value) -> R> Strategy
 mod test {
     use super::*;
 
-    use std::u32;
+    use std::u64;
 
     use crate::strategy::just::Just;
     use crate::test_runner::Config;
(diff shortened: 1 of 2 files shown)


cameron1024:
Hi, thanks for the PR ☺️

This seems reasonable, the main drawback I'm thinking about is the breaking change. This would require proptest 2.0, and I'm not sure if there are any plans for that any time soon.

That said, I can't imagine this would break many people's code in practice. I've only ever used this with an int literal 🤷

kali:
Yeah, same, I would expect most people to use this with the envir or through annotations. But I see the the point, it's strictly a breaking API change.

I've seen there is a merged PR improving the strategy for vectors. It is very possible this PR actually helps also my tensor use case...

cameron1024:
Hopefully that should improve things. That said, I think there's potentially general usefulness in being able to have more than 2^32 iters. If/when we do release a breaking change, this seems like something that should definitely be included. Other than the breaking nature, I can't see any downsides either

matthew-russo:

PR #266, opened 2022-04-18
Update x86 dependency.

gz:
Make sure to replace legacy llvm inline assembly with new asm syntax by updating the x86 crate to the latest version.

Fixes compilation for `hardware-rng` feature on current rustc versions.


rex-remind101:
fixed https://github.com/proptest-rs/proptest/pull/285 will be in next release

Issue #268, opened 2022-05-26
Is Proptest abandoned?


russelldb:
There are a lot of open PRs, the project's own tests no longer pass (since #258) (TBH I can't even get the tests to compile)

Not to be one of those entitled downstream users, but can @AltSysrq please let us know the state of the project, or if someone should fork/takeover as the canonical crate?

Many thanks in advance

nzeh:
There's an updated fork here: https://github.com/nzeh/proptest.  I don't really have the intention of taking over maintainership of this crate.  I'll submit a pull request shortly and would prefer that my changes get incorporated into the main repo.

My fork addresses two issues, alas via (hopefully minor) breaking changes:

  - The tests in `proptest` and `proptest-derive` all pass again.  This came at the cost of removing half-bounded ranges as strategies for `f32` and `f64`.  Those were the failing tests for `proptest`.  After looking at how the `rand` crate generates floats in a range and how `proptest` tries to use that generator, I concluded that there is no way to make the tests pass.  Moreover, the tests are fairly representative of what should be a common use of half-bounded float ranges as strategies, so they would fail also in real-world use and simply shouldn't be supported.

  - In my opinion, the current shrinking strategy for values produced by `prop_flat_map` is almost always a poor choice.  It first tries to shrink the inner value and only when this fails, shrinks the outer value on which the inner value depends.  I ran into this when trying to generate a graph with a number of vertices and a set of vectors that store attributes, one per vertex.  The most natural way to do this is to generate the number of vertices and then `prop_flat_map` this to a strategy that generates a graph with that number of vertices and a set of vectors, each of this length.  Shrinking is completely ineffective here because it will try tons of possibilities of messing with the vector entries and with the edge set of the graph before trying to shrink the number of vertices.  As a result, shrinking will either take hours or will fail to produce a meaningfully shrunk input, depending on the bound on the number of shrinks.

[...]

Issue #181, opened 2020-03-01
Alternative approaches to `prop_flat_map` due to shrinking behavior?


Andlon:
I'm using `proptest` to generate random matrices (dense and sparse). I do so as follows:

1. Use a strategy of `(rows, cols)`.
2. `prop_flat_map` on this strategy and generate a vector of `rows * cols` number of items.

Here's a concrete example:

```rust
pub fn dense_matrix_strategy<T, S>(
    rows: impl Strategy<Value = usize>,
    cols: impl Strategy<Value = usize>,
    strategy: S,
) -> impl Strategy<Value = MockDenseMatrix<T>>
where
    T: Debug,
    S: Clone + Strategy<Value = T>,
{
    (rows, cols).prop_flat_map(move |(r, c)| {
        proptest::collection::vec(strategy.clone(), r * c)
            .prop_map(move |data| MockDenseMatrix::from_row_major(r, c, data))
    })
}
```

This is simple and works very well. However, if I understand correctly, the shrinking procedure for `flat_map` prefers to try to shrink the "inner" strategy (the set of elements) before it tries to shrink the "outer" strategy (rows/cols). This is problematic, because often `proptest` never tries to reduce the dimension of the matrices, instead favoring trying to reduce the size of the individual elements. If I increase `PROPTEST_MAX_SHRINK_ITERS`, it will typically manage to reduce the size as well, but this obviously takes longer to run and I guess there are no guarantees.

I have so far been able to think of a good way around this. I suppose in principle, one could generate a large number of elements first, flat map on this and take only `rows * cols` items from this set, but this seems highly inefficient and you'd need to set a maximum number of elements and it would require some additional care in choosing the rows and cols properly. Does anyone have an idea for how I can approach this problem differently? Any help would be greatly appreciated!

Andlon:
Upon further reflection, it seems to me that these kind of data structures warrant custom `ValueTree` implementations. In particular, we probably want to try to shrink by selecting submatrices (subject to provided constraints, perhaps), which is something that we'll probably have to code ourselves rather than rely on combinators. Am I right in thinking so?

Centril:
[...]

PR #269, opened 2022-07-28
Make tests pass and improve shrinking for prop_flat_map

nzeh:
This pull request aims to address issues #268 and #181, alas by introducing some potentially but hopefully minor breaking changes.  I'm more than happy to discuss whether there's a better way to implement these improvements, without breaking anything or at least breaking less.

The tests in `proptest` and `proptest-derive` all pass again. This came at the cost of removing half-bounded ranges as strategies for `f32` and `f64`. Those were the failing tests for `proptest`. After looking at how the `rand` crate generates floats in a range and how `proptest` tries to use that generator, I concluded that there is no way to make the tests pass. Moreover, the tests are fairly representative of what should be a common use of half-bounded float ranges as strategies, so they would fail also in real-world use and simply shouldn't be supported.

In my opinion, the current shrinking strategy for values produced by `prop_flat_map` is almost always a poor choice. It first tries to shrink the inner value and only when this fails, shrinks the outer value on which the inner value depends. I ran into this when trying to generate a graph with a number of vertices and a set of vectors that store attributes, one per vertex. The most natural way to do this is to generate the number of vertices and then `prop_flat_map` this to a strategy that generates a graph with that number of vertices and a set of vectors, each of this length. Shrinking is completely ineffective here because it will try tons of possibilities of messing with the vector entries and with the edge set of the graph before trying to shrink the number of vertices. As a result, shrinking will either take hours or will fail to produce a meaningfully shrunk input, depending on the bound on the number of shrinks.

I reimplemented the shrinking strategy for `FlattenValueTree` so it shrinks the outer value first and only when that fails, refines the result by shrinking the inner value. Unfortunately, given the API in the `Strategy` trait, I didn't see a way to do this without requiring that the strategy produced by the closure given to `prop_flat_map` implement `Clone`, a requirement that wasn't there before. This may break some custom strategies, but I expect it to break less code than trying to mess with the `Strategy` trait itself and, at least in my use cases, seems like a small price to pay for getting a useful shrinking strategy from `prop_flat_map`.

As I said above, I expect some objections to these changes and am more than happy to discuss whether this can be done better.  However, I think the change to the shrinking strategy of `FlattenValueTree` is crucial, only its implementation may be suboptimal.

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -1,3 +1,20 @@
+## Unreleased
+
+### Breaking Changes
+
+- `prop_flat_map` can only be used with closures that produce a strategy that
+  implements `Copy`.
+
+- Half-bounded ranges (`x..`, `..x`, and `..=x`) are no longer supported for
+  `f32` and `f64`.
+
+### New Features
+
+- The shrink strategy for `prop_flat_map` now shrinks the "outer" strategy
+  first (the one that produces the values given to `prop_flat_map`) and
+  shrinks the "inner" strategy" (the one produced by `prop_flat_map`) only
+  when the outer strategy cannot be shrunk further.
+
 ## 1.0.0
 
 ### Breaking Changes
--- a/proptest/src/strategy/flatten.rs
+++ b/proptest/src/strategy/flatten.rs
@@ -8,11 +8,11 @@
 // except according to those terms.
 
 use crate::std_facade::{fmt, Arc};
-use core::mem;
 
 use crate::strategy::fuse::Fuse;
 use crate::strategy::traits::*;
 use crate::test_runner::*;
+use std::mem;
 
 /// Adaptor that flattens a `Strategy` which produces other `Strategy`s into a
 /// `Strategy` that picks one of those strategies and then picks values from
@@ -33,6 +33,7 @@ impl<S: Strategy> Flatten<S> {
 impl<S: Strategy> Strategy for Flatten<S>
 where
     S::Value: Strategy,
+    <S::Value as Strategy>::Tree: Clone,
 {
     type Tree = FlattenValueTree<S::Tree>;
     type Value = <S::Value as Strategy>::Value;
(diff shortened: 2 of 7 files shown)


proptest/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+## Unreleased
+
+### Breaking Changes
+
+- `prop_flat_map` can only be used with closures that produce a strategy that
+  implements `Copy`.
```
rex-remind101, line 6: I'm concerned about this, I often use Strategies over String, sometimes even `Just(String)`, which is `Clone` but not `Copy`.

Andlon: Is that perhaps a typo? The initial issue mentions `Clone`.

Regardless, it seems a little risky to require `Clone` as it *might* permanently break lots of proptests out there. Would it make sense to instead provide it under a new name, preserving the current `prop_flat_map`? Not ideal either, I know...

I otherwise am happy to see some effort in trying to do something about the shrinking behavior for `prop_flat_map`. Like @nzeh I've run into the issue of poor shrinking many times.

nzeh: Ugh.  Yes, that's a typo.  The updated implementation in the pull request requires `Clone`able strategies, not ones that implement `Copy`.  I aggree with @Andlon that adding even the requirement that the strategy implement `Clone` is suboptimal.  After spending quite some time thinking about how to avoid it at the time, I concluded that doing what my shrink implementation does (shrink the outer strategy first and then the inner one) is not possible with the current architecture of the `ValueTree` trait.  I *think* it is possible if we completely overhaul the architecture of at least parts of proptest, but that was not something I intended to undertake at the time.  Moreover, this overhaul in itself could lead to breakage of code that uses proptest because it would involve changing at least the `ValueTree` trait.

proptest/src/strategy/flatten.rs
```diff
@@ -177,17 +162,20 @@ where
...
-        if self.current.complicate() {
-            return true;
-        } else if self.meta.complicate() {
+        if self.meta.complicate() {
             if let Ok(v) = self.meta.current().new_tree(&mut self.runner) {
-                self.complicate_regen_remaining = self.runner.config().cases;
                 self.current = Fuse::new(v);
+                self.complicate_regen_remaining = self.runner.config().cases;
                 return true;
+            } else {
```
fenollp, line 170: Dead code?

matthew-russo:
Sorry for the delay in getting to this. I think the overall suggestion is sound but I don't think a breaking change is the right approach at the moment. We already have 3 APIs for flat mapping ([`prop_flat_map`](https://docs.rs/proptest/latest/proptest/strategy/trait.Strategy.html#method.prop_flat_map), [`prop_ind_flat_map`](https://docs.rs/proptest/latest/proptest/strategy/trait.Strategy.html#method.prop_ind_flat_map), [`prop_ind_flat_map2`](https://docs.rs/proptest/latest/proptest/strategy/trait.Strategy.html#method.prop_ind_flat_map2)) where the only difference is shrinking behavior so I think in the short term we should add this functionality under a new API and whenever we're ready for a 2.0 release we can decide if a change in the Strategy APIs is the right direction.

I'm also interested in seeing if there's any way to change the behavior without the Clone bound. I'm not opposed to changing the underlying shrinking strategy of `prop_flat_map` if we can do it without an API signature change. I'll try to give the code a closer look this weekend and see if I can think of anything

matthew-russo:

PR #270, opened 2022-08-31
Fix WebAssembly support

### LINKED ISSUE #137: Support building on webassembly
It would be great if proptest could be built on WASM. Slightly related to #47 but more specific, since proptest can almost be compiled to webassembly already.

I'm already disabling [the default forking feature](https://altsysrq.github.io/proptest-book/proptest/forking.html), to avoid the `rusty-fork` dependency.

The only thing blocking building proptest on webassembly is [these few lines](https://github.com/AltSysrq/proptest/blob/d1c4b049337d2f75dd6f49a095115f7c532e5129/proptest/src/arbitrary/_std/env.rs#L89), importing `std::os::unix` inside a `#[cfg(not(target_os = "windows"))]` block.

Could the `osstring_invalid_string` function be constrained more, by excluding `wasm32` for example? Or maybe there's a better solution to avoid using `std::os::unix` on all non-windows targets?

RReverser:
Hides a few paths, that fail at runtime on wasm32-unknown-unknown, under conditional compilation.

Fixes #137.

### DIFF
--- a/book/src/proptest/wasm.md
+++ b/book/src/proptest/wasm.md
@@ -19,6 +19,4 @@ features = ["std"]
 A few APIs are unavailable on `wasm` targets (beyond those which are removed by
 deselecting certain default features):
 
-- Numeric strategies for `i128` and `u128`.
-
 - The `Arbitrary` implementation for `std::env::VarError`.
--- a/proptest/src/arbitrary/primitives.rs
+++ b/proptest/src/arbitrary/primitives.rs
@@ -12,15 +12,10 @@
 use crate::bool;
 use crate::char;
 use crate::num::{
-    f32, f64, i16, i32, i64, i8, isize, u16, u32, u64, u8, usize,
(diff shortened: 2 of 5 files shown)


RReverser:
@matthew-russo Why was this closed?

RReverser:
Oh no, looks like you deleted the master branch instead of renaming in Github settings? That does automatically close all the associated PRs :(

matthew-russo:
fun

matthew-russo:
let me see if i can undo

RReverser:
From personal past experience, I don't think you can. But let me know if you want to cherry-pick this manually or for me to reopen the PR.

matthew-russo:
I'll deal with it -- sorry about the hassle. On the bright side it'll force me to look at all the PRs that had been open without getting attention.

matthew-russo:
this was rebased here:
https://github.com/proptest-rs/proptest/pull/458

which is now replaced targeting the new branch:
https://github.com/proptest-rs/proptest/pull/519

PR #279, opened 2022-11-08
const generic support

cameron1024:
This PR adds `Arbitrary` impls for arrays using const generics. AFAICT, this is not a breaking change.

A few things to note:
 - This PR adds a `uniform` function as the const-generic counterpart to `uniform1`, `uniform2`, etc. I considered making the old ones `#[deprecated]`, but eventually decided against it, since there may be people who don't have access to a recent enough compiler
 - I've added `unarray` as a dependency (a small crate I maintain that provides `core::array::from_fn` on compiler versions before `1.63`. To me, this seems preferable to bumping the MSRV, but others may disagree
 - I've also bumped CI to use Rust 1.55, since that is the earliest version that supports both const generics and the `MaybeUninit` APIs to reimplement `core::iter::from_fn`


rex-remind101 (approved):
Code looks great, I took a look at implementation of the functions from unarray we're using and even though they're `unsafe` they seem fine, fairly simple usage.

Personally, I have no objections to bumping MSVR either, but I'm not sure what versions of rust other users rely on. We may also choose to release a version after we've merged in some chosen set of prs, then bump it for the next version.

cameron1024:
I know that Debian often lags behind a bit ([this page](https://tracker.debian.org/pkg/rustc) suggests they are on 1.62, but I'm not a Debian user so happy to be corrected on this).

IMO, we should establish an MSRV and promise that a bump is a breaking change, but that seems out of scope for this PR. We also might want to wait a bit before doing this so we can get GATs 🤷.

But generally my preference is to keep it as low as possible while adding key features (hence why this PR bumps for const generics, but not for `core::array::from_fn`) 😁

proptest/src/array.rs
```diff
@@ -90,160 +107,111 @@ macro_rules! small_array {
...
-
-            fn new_tree(&self, runner: &mut TestRunner) -> NewTree<Self> {
-                Ok(ArrayValueTree {
-                    tree: [$(self[$ix].new_tree(runner)?,)*],
-                    shrinker: 0,
-                    last_shrinker: None,
-                })
-            }
-        }
+impl<S: Strategy, const N: usize> Strategy for [S; N] {
```
matthew-russo, line 121: do we want to add any test coverage for the const generic versions added?

proptest/Cargo.toml
```diff
@@ -57,6 +57,7 @@ break-dead-code = []
 
 [dependencies]
 bitflags = "1.0.1"
+unarray = "0.1.3"
```
matthew-russo, line 60: i'm going to be a bit nitpicky over this so feel free to push back.

How do you feel about moving this crate in to the proptest organization? Given how lean our dependencies are, I have a mild concern on taking this dependency given we wouldn't have any organizational control over updating it in the event we want or need to. It would also give us a chance to formally review the code as an org and make it a bit more public that we've done our due diligence with the unsafe portions of the code.

embedding a couple comments on the `unarray` code from a quick look through:

- `build_array_result` leaks if an error is encountered, even if it doesn't panic. A written MaybeUninit doesn't have its destructor run without a call to `assume_init` but we just return from the function without ever assuming any of the elements initialized
  - https://github.com/cameron1024/unarray/blob/92123dcb3bde1e4efd313da6fc0df0968d97b7a5/src/build.rs#L53
- same thing goes for `build_array_option` that return `None`
- same thing goes for `map_result` that returns `Err`
- same thing goes for `map_option` that returns `None`
- same thing goes for `from_iter` when the iter sizes don't match
- we could beef test coverage around `build_*` methods to include functions that return `Err` and `None`
- i saw a couple spots where you had `#[cfg_attr(miri, ignore)]`, so i'm assuming you've run this through miri?
- other than that, the unsafe code looked fine

proptest/src/array.rs
```diff
@@ -80,8 +80,25 @@ pub struct ArrayValueTree<T> {
     last_shrinker: Option<usize>,
 }
 
+/// Create a strategy to generate fixed-length arrays.
+///
+/// All values within the new strategy are generated using the given
+/// strategy. The length of the array corresponds to the suffix of the
+/// name of this function.
```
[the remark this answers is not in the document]
cameron1024: Good spot, I just copied the comment from the macro, I'll update

proptest/src/array.rs
cameron1024: We already have a couple of tests that the manual impls used. Are you thinking specifically for the `uniform` function? Or just a test that it works with `[u8; 1000]` for example?

proptest/Cargo.toml
cameron1024: Regarding ownership, I'd be totally happy to move this into the proptest org. A slight bit of context, I usually have `#![deny(unsafe_code)]` in everything I work on, and always put all unsafe code into its own crate to force me to create proper abstractions. These were just some utilities I found myself writing a lot, so grateful to get some more eyes on it.

Regarding the technical issues:
 - the leaking thing sounds pretty bad, I'll definitely fix that. Initially when I read your comment I misread and thought you were talking about panicking, which seems harder to avoid in the generic case. But we can definitely track how many are initialized and drop them in the error cases
 - increasing test coverage sounds like a good idea
 - ironically, the `#[cfg_attr(miri, ignore)]`s are all for proptests :laughing: There's actually an open issue here for running proptests with miri, and I've been meaning to have a look, because IMO they should work well together (assuming performance doesn't become an issue).

But yeah, I'll push a new branch with some changes to `unarray`, and if we want to bring it into the org we can

proptest/src/array.rs
matthew-russo: just a little test that it works. i wouldn't block an approval on it but seems like a decent sanity check

proptest/Cargo.toml
matthew-russo: i don't feel super strongly about the ownership part -- mostly posed it as a straw-man proposal to get some input. Mainly just want to make sure we add this in a way that still feels "open" for consumers of proptest. Given its a new/relatively unknown crate developed by one of the maintainers and the maintainers are new, I don't want it to come off as self-promotion or anything like that (not saying that's what you're doing, i appreciate you willing to share the code and provide a solution that doesn't require a bump in MSRV).

matthew-russo: if we've gotten a few eyes on the crate and don't want to deal with any more overhead we can just move forward as is. If ever needed for maintenance in the future we could always pull parts of it in to proptest without any invasive changes

matthew-russo (approved):

proptest/Cargo.toml
cameron1024: Makes sense. How does this sound as a plan:
 - I'll push a fix to `unarray` for the leak issue you mentioned as well as some better test coverage. The fix I have in mind will fix the egregious cases, but won't cover leaking on panic (last time I had a look at this, it seems like this is not really solvable in the general case anyways)
 - we depend on the fixed version without doing any sort of ownership change for now
 - in the future, if we feel like we want to, we can chat about whether it makes sense to move it into the org (or just certain functions, maybe reimplement, change MSRV, etc, there's quite a few possibilities)

Does that sound good?

matthew-russo: that sounds great! agreed we shouldn't worry about panics

cameron1024: I've got a PR here: https://github.com/cameron1024/unarray/pull/1/files . Feel free to have a look over the implementation. If you're happy with it I'll publish it as 0.1.4 and bump the version that `proptest` depends on

cameron1024: I've bumped the version, let me know if you have any issues with it. AFAICT, it should be pretty solid now. Tested with miri, and added some tests to catch regressions for the leak issue

matthew-russo (approved):

PR #281, opened 2022-11-16
Simplify non_zero arbitrary impl

mikerite:


cameron1024 (approved):
Looks good, should have thought of that 🤦

matthew-russo (approved):

cameron1024:
I think that could work well for unsigned types. 

As for signed types, I think we need to be careful about uniformity. IMO, `NonZeroU64` should be equally likely to produce a 1 or a -1, but naively using `prop_oneof` would leave negative numbers slightly more common than they should be AFAICT. I think we would need to calculate the weights, but that would be tricky since `prop_oneof` expects weights to be a `u32`. For the case of `NonZeroU64` (assuming I've done the maths right), we would need the weights to be `2^63` for the negative numbers and `2^63 - 1` for the positive numbers, both of which overflow a `u32`. And we couldn't simplify this, since `2^63 - 1` is prime.

It also looks like this `u32` is part of the public API, so we can't easily change it. We *may* be able to replace it with a type parameter with `u32` as a default (maybe `N: Into(u128)`), but that would potentially break type inference. [This RFC](https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-generalizing-to-generics) has some guidance, but I don't think this issue is as simple as it might seem.

Or maybe I'm just getting hung up over uniformity. To me, it feels important, but maybe others disagree?

PR #283, opened 2022-11-24
Improve float range sampling

mikerite:
The current implementation has the flaw that it doesn't sample all the floating point values in the range. For example sampling from the `f32` range [0.0, 16777216.0] will never generate any value with a fractional part even through around most of the floating point values in this range have fractional parts. 

This pull request addresses this by using a new algorithm that doesn't have that issue.


cameron1024 (approved):
Looks good, thanks 🎉

matthew-russo (approved):

PR #285, opened 2022-12-19
[fix-ci-nigtly] 

rex-remind101:
ci was broken on nightly this series of fixes should correct it

* dependencies.x86 was bumped to latest current version. x86 crate does not advertise an MSRV, but this was just broken on even our MSRV because the old x86 version was using a macro removed from rust, so bumping to latest seems fair.
* the calculation for the arbitrary impl of Layout was using a max_size that was too large and overflowing Layout. this has been fixed.
* test for arbitrary AllocError was referering to AllocErr which does not exist, this was fixed.
* NoneError has been removed from rust so it was subsequently removed from proptest. It was blocking compilation. evidence: https://github.com/rust-lang/rust/issues/85614
* try_reserve is stable so removed from unstable features
* try_trait has been changed to try_trait_v2 so that was fixed in Cargo.toml.

### DIFF
--- a/proptest-derive/Cargo.toml
+++ b/proptest-derive/Cargo.toml
@@ -27,7 +27,7 @@ proptest = { version = "1.0.0", path = "../proptest" }
 # features. However, due to
 # https://github.com/laumann/compiletest-rs/issues/166, the default features of
 # compiletest-rs fail to compile, but the stable fallback works fine.
-compiletest_rs = { version = "0.3.19", features = ["tmp", "stable"] }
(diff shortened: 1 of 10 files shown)


matthew-russo (approved):

matusf:
Hi, just fyi, this PR can close also https://github.com/proptest-rs/proptest/pull/266

rex-remind101:
fyi, we're still trying to piece together why this last compiletest is failing.

matthew-russo (approved):

PR #287, opened 2023-01-06
Config : support disabling failure persistence via env var

### LINKED ISSUE #276: Add an environment variable to disable failure persistence
As the title says, there should be an easy way to globally disable file failure persistence.

matthew-russo:
closes #276

### DIFF
--- a/proptest/src/test_runner/config.rs
+++ b/proptest/src/test_runner/config.rs
@@ -43,7 +43,10 @@ const FORK: &str = "PROPTEST_FORK";
 const TIMEOUT: &str = "PROPTEST_TIMEOUT";
 #[cfg(feature = "std")]
 const VERBOSE: &str = "PROPTEST_VERBOSE";
+#[cfg(feature = "std")]
 const RNG_ALGORITHM: &str = "PROPTEST_RNG_ALGORITHM";
+#[cfg(feature = "std")]
+const DISABLE_FAILURE_PERSISTENCE: &str = "PROPTEST_DISABLE_FAILURE_PERSISTENCE";
 
 #[cfg(feature = "std")]
 fn contextualize_config(mut result: Config) -> Config {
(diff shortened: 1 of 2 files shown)


proptest/src/test_runner/config.rs
```diff
@@ -43,7 +43,10 @@ const FORK: &str = "PROPTEST_FORK";
 const TIMEOUT: &str = "PROPTEST_TIMEOUT";
 #[cfg(feature = "std")]
 const VERBOSE: &str = "PROPTEST_VERBOSE";
+#[cfg(feature = "std")]
```
matthew-russo, line 46: not strictly needed but noticed it was inconsistent

proptest/src/test_runner/config.rs
```diff
@@ -217,7 +223,9 @@ pub struct Config {
     /// See the docs of [`FileFailurePersistence`](enum.FileFailurePersistence.html)
     /// and [`MapFailurePersistence`](struct.MapFailurePersistence.html) for more information.
     ///
-    /// The default cannot currently be overridden by an environment variable.
+    /// You can disable failure persistence with the `PROPTEST_DISABLE_FAILURE_PERSISTENCE`
+    /// environment variable but its not currently possible to set the persistence file
+    /// with an environment variable.
```
cameron1024, line 228: There was recently a PR adding a comment to these saying "the env var only has an effect when the `std` feature is enabled". I'm assuming that applies here? If so, should we add that warning?

cameron1024 (approved):
One minor thing, otherwise LGTM

proptest/src/test_runner/config.rs
matthew-russo: Updated

rex-remind101 (approved):
looks good 👍

PR #288, opened 2023-01-06
Const generic arbitrary

cameron1024:
In the previous ticket, I added `Strategy` impls for const generic arrays of strategies, but forgot about `Arbitrary` :facepalm: 

This one adds `Arbitrary` impls for const generic arrays of types which impl `Arbitrary`.

As with the previous PR, I don't think this is a breaking change.

### DIFF
--- a/proptest/src/arbitrary/arrays.rs
+++ b/proptest/src/arbitrary/arrays.rs
@@ -25,14 +25,24 @@ macro_rules! array {
     )* };
 }
 
-array!(
-    1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21,
-    22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32
-);
+impl<A: Arbitrary, const N: usize> Arbitrary for [A; N] {
+    type Parameters = A::Parameters;
+    type Strategy = UniformArrayStrategy<A::Strategy, [A; N]>;
+
+    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+        let base = any_with::<A>(args);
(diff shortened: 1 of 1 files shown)


rex-remind101 (approved):
just to make sure i'm understanding, this change makes it so that `Arbitrary` is inherited across the board for const generic arrays so the user doesn't need to use a macro?

if so seems good to me. one thing i've noticed though is the macro `arbitrary!` seems to be used for these declarations internally instead. however, it seems pretty simple to just inherit the trait directly like you have here so i don't have a problem with that.

cameron1024:
I don't think that's quite right:

The `array` macro only serves to write boilerplate impls.

It's roughly equivalent to :
```rust
impl<A: Arbitrary> Arbitrary for [A; 1] { ... }
impl<A: Arbitrary> Arbitrary for [A; 2] { ... }
// ...
impl<A: Arbitrary> Arbitrary for [A; 32] { ... }
```
except with less typing. `array!` was never exposed to users.

The new code uses const generics to provide impls for `[A; N]`, regardless the value of `N`.

Another way of thinking about it is that the previous code was a bit like this hypothetical syntax:

```rust
impl<A: Arbitrary, const N: usize> Arbitrary for [A; N] where A >= 1, A <= 32 {}
```
and this PR just removes that `where` clause. Of course this isn't supported at the language level, but conceptually we're just relaxing a constraint, so it shouldn't break compat

As for `arbitrary!`, I haven't looked at it in great detail, but I generally find looking at decl-macros that I didn't write pretty tricky, and avoid them where possible. IMO, this impl is pretty self-describing and readable

rex-remind101:
Thanks for the explanation, that all makes good sense to me :)

matthew-russo (approved):
looks good to me

Issue #290, opened 2023-01-12
Testing a small number of cases across a large number of CI runs


adeschamps:
I have some tests that are a little slow. Some are integration tests that take on the order of a few seconds, although even tests that take ~10ms still add up when many iterations are tested.

During development, I like to explore as much of the input space as possible, and I'm happy to set `PROPTEST_CASES` to a large number and let it run in the background. But once I'm satisfied with my code, my priorities shift to not slowing down CI too much. Since CI runs frequently, could I configure it to test just a few cases and still expect that over the course of many CI runs, the input space will still be explored?

Put another way, if I run a single test with the default value of `PROPTEST_CASES=256`, compared to 256 CI runs where `PROPTEST_CASES=1`, can I expect similar coverage?

rex-remind101:
I think this is a harder question to provide a complete answer for than it seems on the surface. If your test is over a function those interface is just scalar arguments, then default strategies will randomly sample within consistent bounds of those scalar arguments. In that case, coverage of 1 test (run * 256 cases) == (256 test runs * 1 case). With more complex testing strategies and implementations that may have nonlinear characteristics (such as where there's memoization is included), I'm not those runs will be exactly equal random variables.

rex-remind101:
For full disclosure I am a new maintainer here so I'll need to think about this more, there may be a simple answer for all cases and not just basic scalar cases. If someone else has that answer already feel free to provide it.

adeschamps:
Thanks for the response! In my tests I don't have any memoization or state -- it's all scalars and collections.

If I'm reading the code correctly (and also based on my observations when running the tests) it seems like one thing to be aware of is the saved seeds from a regressions file. Since the runner runs those first, and they count towards the number of cases run, I believe that if I were to set `PROPTEST_CASES <= num_regressions` then the runner would never explore new inputs. This is something I can work around -- for example, by not committing the regressions file, and instead converting regressions into separate unit tests.

rex-remind101:
> If I'm reading the code correctly (and also based on my observations when running the tests) it seems like one thing to be aware of is the saved seeds from a regressions file.

This seems like a bug given this.
https://github.com/proptest-rs/proptest/blob/1ad340e7c2276d0afc4c0ac13040a1861d198895/proptest/src/test_runner/config.rs#L182-L186

[...]

Issue #284, opened 2022-12-05
Provide a way to also supply multiple manual test cases, that is, parameterized or explicit cases.


Clockwork-Muse:
Often when writing tests, it's good practice to include specific known test cases (usually around boundaries).

Currently, there seems to be these options:
 1. Manually write an individual test for each case.  This works with one or a small number of boundaries, but becomes burdensome for any large number of similarly-structured cases. 
 2. Use a macro to generate cases (or a crate that provides such a macro).  This works, but generates additional methods under the covers.  Failures refer to the failing line in the duplicated method, but picking the case out of the list may be hard in some cases.  Too, this requires an additional crate, and almost certainly one that doesn't work with proptest.
 3. With proptest, use `sample` or `uniform` to select from a list of cases.  At lower counts of cases, this should run all of them.  However, this is really a misuse of those methods, and may end up confusing the engine besides.  Additionally, it can be hard to spot which test case is failing with large manual lists.
  (4. Might be possible to add manual cases to the regression files?  Haven't tried, but this seems like even more of a hack, and that you're more likely to lose cases when tests change)

What I'd like is some way to provide explicit, multiple, test cases, in addition to the ones generated by proptest.  _All_ cases should be run (as opposed to only being maybe run, as with `sample`/`uniform`).  Simplification scenarios would likely need to ignore the cases (especially for composed strategies), but being able to consider them as "known good" data would be a bonus.  Ideally, it would also be possible to trace back to which manual case failed (eg, a line number, in addition to the data).

Andlon:
I would like to chime in and say that this is one of the things I've been missing the most from `proptest` as well. Sometimes I find myself writing manual test cases *with the same test logic* exactly as described here, which seems highly redundant and a waste of time.

rex-remind101:
To try and paraphrase, rather than always having random variables as input to a proptest, you'd like to be able to also run through a set of constant inputs representing specific example test cases? Assuming I'm understanding correctly this sounds beneficial. Do you have a straw man api for how you'd like this to look?

Clockwork-Muse:
Yes, the idea is to have a set of constant inputs to run against tests, preferably in addition to any random inputs.

As for strawman...

Most of the other test frameworks I'm familiar with in other languages have you decorate the method directly (take C#'s xunit as an example), and usually allow supplying both multiple "simple" cases as well as generators.  So something like:
```rust
proptest! {
    #[test]
    #[case(5, 7)]
[...]

PR #295, opened 2023-01-25
[replays-dont-count-cases] 

rex-remind101:
replays/persisted failures are not counted against successful cases.

this way when `PROPTEST_CASES` <= `number of persisted cases` we will not end up never creating new cases to test against. rather, all persisted cases will be tested _and then_ a new number of cases will run equal to the number of `PROPTEST_CASES`.

addresses https://github.com/proptest-rs/proptest/issues/290

### DIFF
--- a/proptest/src/test_runner/errors.rs
+++ b/proptest/src/test_runner/errors.rs
@@ -33,8 +33,32 @@ pub enum TestCaseError {
     Fail(Reason),
 }
 
+/// Indicates the type of test that ran successfully.
+///
+/// This is used for managing whether or not a success is counted against
+/// configured `PROPTEST_CASES`; only `NewCases` shall be counted.
+///
+/// TODO-v2: Ideally `TestCaseResult = Result<TestCaseOk, TestCaseError>`
+/// however this breaks source compability in version 1.x.x because
+/// `TestCaseResult` is public.
+#[derive(Debug, Clone)]
+pub(crate) enum TestCaseOk {
+    NewCaseSuccess,
+    PersistedCaseSuccess,
+    ReplayFromForkSuccess,
+    CacheHitSuccess,
+    Reject,
+}
+
 /// Convenience for the type returned by test cases.
 pub type TestCaseResult = Result<(), TestCaseError>;
+
+/// Intended to replace `TestCaseResult` in v2.
+///
+/// TODO-v2: Ideally `TestCaseResult = Result<TestCaseOk, TestCaseError>`
+/// however this breaks source compability in version 1.x.x because
+/// `TestCaseResult` is public.
+pub(crate) type TestCaseResultV2 = Result<TestCaseOk, TestCaseError>;
 
 impl TestCaseError {
     /// Rejects the generated test input as invalid for this test case. This
--- a/proptest/src/test_runner/runner.rs
+++ b/proptest/src/test_runner/runner.rs
@@ -178,27 +178,34 @@ fn call_test<V, F, R>(
     _runner: &mut TestRunner,
     case: V,
     test: &F,
-    replay: &mut R,
+    replay_from_fork: &mut R,
     result_cache: &mut dyn ResultCache,
     _: &mut ForkOutput,
-) -> TestCaseResult
+    is_from_persisted_seed: bool,
+) -> TestCaseResultV2
 where
     V: fmt::Debug,
     F: Fn(V) -> TestCaseResult,
     R: Iterator<Item = TestCaseResult>,
 {
-    if let Some(result) = replay.next() {
-        return result;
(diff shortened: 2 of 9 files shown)


proptest/src/test_runner/errors.rs
```diff
@@ -33,9 +33,32 @@ pub enum TestCaseError {
...
+/// TODO-v2: Ideally `TestCaseResult = Result<TestCaseOk, TestCaseError>`
+/// however this breaks source compability in version 1.x.x because
+/// `TestCaseResult` is public.
+#[derive(Debug, Clone)]
+pub(crate) enum TestCaseOk {
+    NewCaseSuccess,
+    ReplaySuccess,
+    CacheHitSuccess,
+    Reject,
+}
```
matthew-russo, line 50: Not expecting any changes inline to this PR as these thoughts aren't fully baked but wanted to run some stuff by you and see how you're thinking of these things. Also since this is a crate-private type, we can evolve this independently of crate version so that gives us some flexibility for the future.

I've been thinking of how to add some new functionality to support #284 (providing manual, explicit cases) and was thinking of introducing a new wrapper type for cases that carry context on their source so we can have something like:
```
enum Case<T: fmt::Debug, S: Strategy<Value=T>> {
    UserProvided(T),    
    Generated(S),
    Persisted(S, Seed),
    Replay(S, TestCaseResult),
}
```

or something along these lines and then the test-runner code can branch on these accordingly to affect seed, success counts, whether we actually run the test or just emulate it based on replay value, etc.

We could be left with a result type of:
```
enum TestCaseOk {
    Success,
    CacheHit,
    Reject,
}
```

Not entirely sure all this would end up working out with the current code structure as I haven't actually started implementing it.

proptest/src/test_runner/runner.rs
```diff
@@ -178,38 +178,46 @@ fn call_test<V, F, R>(
...
-        return result.clone();
+        return result.clone().map(|_| TestCaseOk::CacheHitSuccess);
     }
 
     let result = test(case);
     result_cache.put(cache_key, &result);
-    result
+    result.map(|_| {
+        if is_from_persisted_seed {
+            TestCaseOk::ReplaySuccess
```
matthew-russo, line 204: I don't think this is the right return value. we're actually running the test for this result so its not really a "replay" in the context of what "replay" means for forked tests. I commented on the enum def as well but to be explicit, what do you think about a `PersistedCaseSuccess` variant?

proptest/src/test_runner/errors.rs
matthew-russo: And more direct comment on this code after reading some more below, I think we'd want a `PersistedCaseSuccess` variant. Right now it seems `ReplaySuccess` is being overloaded for both persisted tests as well as forked tests.

proptest/src/test_runner/runner.rs
```diff
@@ -548,6 +562,10 @@ impl TestRunner {
             |_| panic!("Ran past the end of the replay"),
             replay.steps.into_iter(),
             ForkOutput::empty(),
+            // Since all of our test runs happened in forks already, we need to
+            // force the runner to count successful replays against our total
+            // success count.
+            true,
```
matthew-russo, line 568: not a huge fan of having boolean flags flip behavior like this. having explicit variants for `PersistedCaseSuccess` and `ReplaySuccess` would remove the need for this right? I think we always want replays to count against the success count and we never want persisted cases to count against success count.

proptest/src/test_runner/runner.rs
rex-remind101: good call out, will make that change 👍

proptest/src/test_runner/errors.rs
rex-remind101: that seems like a good approach to me 👍

proptest/src/test_runner/runner.rs
rex-remind101: updated

matthew-russo (approved):

cameron1024 (approved):
Apologies for taking so long looking into this :sweat_smile: 

Code looks good, one minor thought I have is that it may be surprising to users if you have `PROPTEST_CASES=1` and it still runs multiple tests. This is something I do fairly frequently - I have some tests that take a fairly long time to run, so want to only run it once for a shorter local dev cycle.

At the very least, we should document this (though IIRC the new behaviour *is* what the documentation said, if so, then that's probably fine).

Perhaps it's worth adding something like `PROPTEST_TOTAL_CASES` or something, to set an absolute limit on the number of tests run, even if that means skipping known regressions. Though that's for sure out of scope for this PR. I'm just not sure about whether it's something we should block 1.1 on

PR #299, opened 2023-02-06
Remove quick_error dependency

cole-miller:
This removes the direct dependency on the quick-error crate, replacing the single use of `quick_error!` with (brief, straightforward) manual trait impls. quick-error remains a *transitive* dependency because it's pulled in by rusty-fork.

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -23,6 +23,10 @@
 - Add `Arbitrary` impls for `core::num::NonZero*`
 - Adds ability to disable failure persistence via env var `PROPTEST_DISABLE_FAILURE_PERSISTENCE`
 
+### Other Notes
+
+- `proptest` no longer depends on the `quick-error` crate.
+
 ## 1.0.0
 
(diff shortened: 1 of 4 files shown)


matthew-russo (approved):
lgtm. just for my own understanding the reason to remove this is just to make proptest slightly leaner correct? there weren't any underlying issues with `quick_error`?

proptest/CHANGELOG.md
```diff
@@ -23,6 +23,10 @@
 - Add `Arbitrary` impls for `core::num::NonZero*`
 - Adds ability to disable failure persistence via env var `PROPTEST_DISABLE_FAILURE_PERSISTENCE`
 
+### Other Notes
+
+- `proptest` no longer depends on the `quick-error` crate.
```
cameron1024, line 28: super minor nit, could this say "directly depends" - I think it's worth making the distinction here, since it *does* still depend on it (via rusty-fork)

cameron1024 (approved):
Looks good, thanks :tada:

PR #301, opened 2023-02-08
add msrv to readme

cameron1024:
Adds a section to the readme detailing our MSRV policy.

Perhaps it's worth leaving this PR open for a while if community members want to express their opinions

### DIFF
--- a/proptest/README.md
+++ b/proptest/README.md
@@ -27,6 +27,12 @@ sees passive maintenance.
 See the [changelog](https://github.com/AltSysrq/proptest/blob/master/proptest/CHANGELOG.md)
 for a full list of substantial historical changes, breaking and otherwise.
 
+### MSRV
+
+The current MSRV of this crate is 1.60.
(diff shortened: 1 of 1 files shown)


proptest/README.md
```diff
@@ -27,6 +27,12 @@ sees passive maintenance.
 See the [changelog](https://github.com/AltSysrq/proptest/blob/master/proptest/CHANGELOG.md)
 for a full list of substantial historical changes, breaking and otherwise.
 
+### MSRV
+
+The current MSRV of this crate is 1.60.
```
matthew-russo, line 32: we can just put it in the cargo file:
https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-rust-version-field
and not worry about needing to update the readme

matthew-russo (approved):

proptest/README.md
rex-remind101: made that change here https://github.com/proptest-rs/proptest/pull/307

PR #306, opened 2023-03-20
fix nightly derive tests

rex-remind101:
nightly needs to be cleaned before running derive tests otherwise it will fail with
```
unexpected error: '10:1: 10:30: multiple candidates for `dylib`
dependency `proptest_derive` found [E0464]'
```
and other errors


cameron1024 (approved):
Thanks :)

PR #308, opened 2023-03-25
Update syn and friends to v1

CosmicHorrorDev:
I was curious why `syn` v0.15 still has so many downloads. I think this may be the main culprit :grin:

This updates all of the proc_macro related bits to v1. I'm planning on updating `syn` to v2, but figured breaking things into two PRs would make reviewing easier. The changes are largely:

- Handling `Meta::Word(_)` -> `Meta::Path(_)` since this was changed to take arbitrary paths instead of a single ident
  - Fix: Ensure that the path is just a single ident
- `IntSuffix` got replaced with just returning a `&str` for the suffix
  - Fix: Switch all the logic to just use the `&str`. Could also make our own version of `IntSuffix`, but wasn't sure if that would be worth it
- `Tuple` and `Tuple2` were causing a lot of headaches
  - Fix: Remove them since they seem unused?

### DIFF
--- a/proptest-derive/src/interp.rs
+++ b/proptest-derive/src/interp.rs
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use syn::{self, BinOp as B, Expr as E, IntSuffix as IS, Lit as L, UnOp as U};
+use syn::{self, BinOp as B, Expr as E, Lit as L, UnOp as U};
 
 /// Adapted from https://docs.rs/syn/0.14.2/src/syn/lit.rs.html#943 to accept
 /// u128.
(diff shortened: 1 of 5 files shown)


proptest-derive/src/interp.rs
```diff
@@ -69,62 +69,53 @@ fn parse_lit_int(mut s: &str) -> Option<u128> {
...
+            "i8" if val <= i8::MAX as u128 => val,
+            "i16" if val <= i16::MAX as u128 => val,
+            "i32" if val <= i32::MAX as u128 => val,
+            "i64" if val <= i64::MAX as u128 => val,
+            "u8" if val <= u128::from(u8::MAX) => val,
+            "u16" if val <= u128::from(u16::MAX) => val,
+            "u32" if val <= u128::from(u32::MAX) => val,
+            "u64" if val <= u128::from(u64::MAX) => val,
+            "usize" if val <= usize::MAX as u128 => val,
+            "isize" if val <= isize::MAX as u128 => val,
```
CosmicHorrorDev, line 96: Wasn't sure if these were intentionally being treated as `(i|u)128` for a reason, or if it was a copy-paste style error

matthew-russo (approved):
lgtm. thanks for doing this!

cameron1024 (approved):
Thanks :grin:

messense:
Can we have a new `proptest-derive` crates.io release please? Thank you!

cameron1024:
We can take a look. What specific features are missing from the latest crates.io release

PR #257, opened 2022-01-03
[WIP] alternative state machine support

tzemanovic:
Thank you for the initial state machine tests from #84! I took some inspiration from it and re-used the example from it but I used a bit different approach for the implementation. This version is mostly inspired by the Erlang's eqc_statem (see Finding Race Conditions in Erlang with QuickCheck and PULSE) with some key differences, which are not yet documented. This version only supports sequential state machine for now, but it should be possible to add concurrent execution using the linearization technique employed in eqc_statem

PR #310, opened 2023-03-26
State machine testing

tzemanovic:
This is a new version of https://github.com/proptest-rs/proptest/pull/257 with some refinements.

Accompanying blog post (it contains the guide to be included in the Proptest book): https://tzemanovic.gitlab.io/posts/state-machine-testing-with-proptest/

A summary of differences from #257:
- `StateMachineTest::init_test` takes `abstract_state` by immutable reference instead of an owned value.
- `StateMachineTest::apply_concrete` and `StateMachineTest::invariants` additionally take `abstract_state` immutable reference, which can be used for comparison with the `ConcreteState`.
- The `StateMachineTest::invariants` are also checked on the initial state
- `StateMachineTest::test_sequential` now takes `proptest::test_runner::Config`, which can be used enable printing of the transitions before they're applied in the test with `verbose = 1` or higher to make it easier to follow what the state machine does and to aid in diagnostics. Because of this change, the support for state machine test setup without the `prop_state_machine!` macro has been dropped as it's not possible to access the config value without it.
- Improved the state machine's rustdoc.
- Some internal refactors have been applied in the state machine strategy implementation.
- Added tests for the sequential strategy implementation.
	- Found and fixed an issue in which an index of a rejected simplification wasn't getting cleared from `SequentialVecTree::shrinkable_transitions`, which could cause an infinite loop.
 - Added a new example `state_machine_echo_server.rs`
- Added a Proptest Book page.
- Edit 16/4/23: renamed (thanks to @matthew-russo!):
  - `trait AbstractStateMachine` to `trait ReferenceStateMachine`
  - `type ConcreteState` to `type SystemUnderTest`
  - `type Abstract` to `type Reference`
  - `fn apply_abstract` to `fn apply`
  - `fn apply_concrete` to `fn apply`
  - `fn invariants` to `fn check_invariants`


russelldb:
> StateMachineTest::apply_concrete and StateMachineTest::invariants additionally take abstract_state immutable reference, which can be used for comparison with the ConcreteState.

Yes! Bravo! 👏

fenollp:
We're still using the version over at https://github.com/proptest-rs/proptest/pull/257

Is there any way to run setup/cleanup code per run?

I'd like to delete resources that I create during the execution of my sequence of transitions. I'd like not to put too many things esp effectful code in `impl Drop`.

See for example in an Erlang implementation: https://proper-testing.github.io/apidocs/
```
?SETUP(<Setup_fun>, <Prop>)
    Adds a setup <Setup_fun>ction to the property which will be called before the first test. This function has to return a finalize function of arity 0, which should return the atom ok, that will be called after the last test. It is possible to use multiple ?SETUP macros on the same property.
```

russelldb:
> > The hard part is running code after a whole sequence of calls is executed and before the model struct is being dropped.
> 
> I see - you should be able to do that when you override the default `test_sequential` after the for loop that goes through the transitions. I'm not sure how common this is, but maybe there could be an additional method to allow that, e.g. `tear_down`

I do this _all the time_. Inset set up code before the loop in `test_sequential` and teardown afterwards. Works fine

itamarst:
I am very excited about this.

One suggestion: the heap example is _slightly_ too simple. 

In particular, the key differentiator in my mind of stateful testing vs. non-stateful testing is the ability to generate transitions based on current state. For the heap example, I could just do a normal proptest and generate a sequence of Pop/Push commands.

On the other hand, an example of a key-value map, rather than a heap, where you insert values and might have transition to get _existing_ keys, shows the extra value you get from having a model state. Generating a `Transition::Get(existing_key)` is where the stateful testing really lets you express new kinds of tests. And the echo server  example does that, but with more complexity.

Second suggestion: the docs could probably benefit from a walkthrough of implementing a concrete, simple example (e.g. the key-value map). In the [four-quadrant documentation model](https://documentation.divio.com/) this is missing the Tutorial quadrant.

Neither of these seem like blockers, since they can always be done as a second pass.

proptest/src/test_runner/state_machine.rs
```diff
@@ -0,0 +1,193 @@
...
+// except according to those terms.
+
+//! Test declaration helpers and runners for abstract state machine testing.
+
+use super::Config;
+use crate::std_facade::Vec;
+use crate::strategy::state_machine::AbstractStateMachine;
+
+/// State machine test that relies on an abstract state machine model
+pub trait StateMachineTest {
```
matthew-russo, line 17: do we and if so, how do we want to think about testing async code? have you done any state machine testing with async code and if so, what was the experience like?

matthew-russo (commented):
I'll need a few more passes over this as its quite dense but I like where this is going. I mostly looked over the general structure of things and the traits. I put in some high level questions. Most are naming bikesheds. A few are some high level directional questions.

Things i'll need to review in more detail in subsequent reviews:
- prose/book documentation
- examples
- SequentialValueTree
- macro

fenollp:
Overriding `test_sequential` for the purpose of `teardown` i.e. inserting code means duplicating the rest of the code, at the risk of this duplication getting stale.

I'd much prefer passing a config option or an additional `teardown` trait with a default noop impl. Please.

proptest/src/test_runner/state_machine.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Override this function to add some teardown logic on the concrete state
+    /// at the end of each test case. The default implementation simply drops
+    /// the state.
+    fn teardown(state: Self::ConcreteState) {
+        // This is to avoid `unused_variables` warning
+        let _ = state;
+    }
+
+    /// Run the test sequentially. You typically don't need to override this
+    /// method.
```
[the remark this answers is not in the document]
russelldb: I override is _all_ the time when doing model based testing, as I need to update the model alongside the thing being tested, and this is the place to do it.

matthew-russo (commented):
Still need to work through the `Sequential` and `SequentialValueTree` but other than that things look good. I have one semi-large discussion topic around crates and versioning. Thanks for all the hard work on this

proptest/src/lib.rs
```diff
@@ -96,6 +96,7 @@ pub mod tuple;
 pub mod option;
 pub mod result;
 pub mod sample;
+pub mod state_machine;
```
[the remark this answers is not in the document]
tzemanovic: After more thought, I think this is a good idea. I don't think there should be any issue with separating it, I'll give it a try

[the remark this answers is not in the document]
tzemanovic: There are some symbols that are currently `pub(crate)` that would need to be made public:

- `bits::varsize::VarBitSet`
- `collection::SizeRange::start_end_incl`
- `num::sample_uniform_incl` - this one is a pretty tiny indirection

wdyt?

book/src/proptest/state-machine.md
```diff
@@ -0,0 +1,174 @@
+# State Machine testing
+
+Before using state machine testing, it is recommended to be at least familiar with the basic concepts of Proptest itself as it's built on its essential foundations. That is:
+
+- Strategies are composed and used to generate inputs to a state machine test.
+- A test will attempt to shrink the generated transitions sequence on a discovery of a case that breaks some properties.
+- It will capture regressions file with a seed that can be used to deterministically repeat the found case.
+
+Be cautious that it may not be fully self-explanatory how to use it from its API and there are some opportunities for paper cuts. What follows is a more detailed description that will guide you through it.
```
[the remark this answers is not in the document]
tzemanovic: I removed it in [9cc1530](https://github.com/proptest-rs/proptest/pull/310/commits/9cc15302ab86f069232140312ef63629b080bdd2) with the other book edits

proptest/src/test_runner/runner.rs
```diff
@@ -39,7 +39,9 @@ use crate::test_runner::rng::TestRng;
 const ENV_FORK_FILE: &'static str = "_PROPTEST_FORKFILE";
 
 const ALWAYS: u32 = 0;
-const SHOW_FALURES: u32 = 1;
+/// Verbose level 1 to show failures. In state machine tests this level is used
+/// to print transitions.
+pub const SHOW_FAILURES: u32 = 1;
```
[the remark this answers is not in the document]
tzemanovic: renamed in [5dbaac5](https://github.com/proptest-rs/proptest/pull/310/commits/5dbaac53a407a49fd983cf47af2c69307f810ac5)

proptest/src/lib.rs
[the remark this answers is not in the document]
matthew-russo: > bits::varsize::VarBitSet

this one seems fine to me

> collection::SizeRange::start_end_incl

this one also seems fine

> sample_uniform_incl

the impl of this ones only one line so i'd probably just call that directly rather than exposing it but i dont have strong opinions.
```
Uniform::new_inclusive(start, end).sample(run.rng())
```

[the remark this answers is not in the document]
tzemanovic: alright, there's proptest-state-machine now (done in https://github.com/proptest-rs/proptest/pull/310/files/4178671c9ea77b29f0110cac4ac5a88fc99d2121..a1dc279667001c0cce9bafa5f1b9e3d47bc9cc24)

proptest-state-machine/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+homepage = "https://proptest-rs.github.io/proptest/proptest/state-machine.html"
+documentation = "https://docs.rs/proptest-state-machine/latest/proptest-state-machine/"
+keywords = ["property", "testing", "quickcheck", "fuzz", "state-machine"]
+categories = ["development-tools::testing"]
+description = """
+State machine based testing support for proptest.
+"""
+
+[dependencies]
+proptest = { version = "1.0.0", path = "../proptest" }
```
matthew-russo, line 17: we can't publish to crates with path dependencies:

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies

i think we'll also need to depend on a yet-to-be-released version "1.2.0" which has the changes to make the necessary symbols public.

matthew-russo (approved):
Approving as is so we don't leave this large PR open longer than necessary but I think we'll need a change to the toml file to not use a path dependency and bump version to 1.2.0.

We should also update the changelog for `proptest` with a note of why we're making some things public and start a changelog for `proptest-state-machine`

proptest-state-machine/Cargo.toml
tzemanovic: Thank you for much for your help and the thorough review! 

I think the docs say that we can't publish with a dependency that *only* has a path, but when there's also a `version` field it will use that for crates.io. You're right that the version needs to be bumped for this to work though. Let me add the changelogs, squash the fixup commits and rebase. I think we'll need to do the 1.2.0 release before publishing the state machine crate.

tzemanovic: The bump to 1.2.0 broke the CI for now. The last run before that was successful: https://github.com/proptest-rs/proptest/actions/runs/4870392917/jobs/8686088563. This is the diff: https://github.com/proptest-rs/proptest/compare/0e8a0c004957cb0ea41eb0072c04f26e594f70af..504393b33e4e306e7dc7018aa841e43d9de058b9

proptest-state-machine/src/test_runner.rs
```diff
@@ -0,0 +1,195 @@
...
+///     #[test]
+///     fn run_with_macro(
+///         (initial_state, transitions) in MyTest::sequential_strategy(1..20)
+///     ) {
+///        MyTest::test_sequential(initial_state, transitions)
+///     }
+/// }
+/// ```
+#[macro_export]
+macro_rules! prop_state_machine {
```
cameron1024, line 158: Not really an issue for this PR, but this macro makes me glad that we're not shipping this in `proptest` proper just yet - I have a feeling it will be possible to get a nicer API with the attr-macro

proptest/src/test_runner/runner.rs
```diff
@@ -39,7 +39,9 @@ use crate::test_runner::rng::TestRng;
 const ENV_FORK_FILE: &'static str = "_PROPTEST_FORKFILE";
 
 const ALWAYS: u32 = 0;
-const SHOW_FALURES: u32 = 1;
+/// Verbose level 1 to show failures. In state machine tests this level is used
+/// to print transitions.
+pub const INFO_LOG: u32 = 1;
```
cameron1024, line 44: Is there a reason this needs to be pub? Do the others also not need to be pub?

cameron1024 (approved):
Code all looks good, just a few small comments about some minor details, but otherwise LGTM :tada: 

This looks really cool, thanks for pushing it over the finish line :pray:

itamarst:
Thanks to everyone who worked on this!

book/src/proptest/state-machine.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Extra tips
+
+Because a state machine test may be heavier than regular prop tests, if you're running your tests in a CI you may want to override the default `proptest_config`'s `cases` to include fewer cases in a single run. You can also use `PROPTEST_CASES` environment variable and during development it is preferable to override this to run many cases to get a better chance of catching those pesky ~~bugs~~, erm, defects.
+
+> Given that there are thought to be in the region of another four million species that we have not yet even named, there is no doubt that scientists will be kept happily occupied studying them for millennia, so long as the insects remain to be studied. Would the world not be less rich, less surprising, less wonderful, if these peculiar creatures did not exist?
+>
+> -- <cite>Dave Goulson, Silent Earth</cite>
+
+So let's leave bugs alone and only squash defects instead!
```
rex-remind101, line 146: I know this is trying to be a bit fun and i'm being somewhat of a downer, but i think this interlude about bugs actually detracts from the clarity of the documentation. Everything else I've found to be well written and on topic without being too terse nor too wordy, and then this kind of left my head scratching and I had to re-read it. If you don't want to use the word "bug", that's totally fine, we can use the word "defect" without the need to elaborate on that. Keep in mind too this is lib documentation, not a blog, which people want to scan quickly (meaning less distraction is better) while still getting enough information (and related reference links) they need to do the job.

rex-remind101: > if you're running your tests in a CI you may want to override the default `proptest_config`'s `cases` to include fewer cases in a single run

I'm not sure what's most common, but in my case CI is has significantly more horsepower than dev's laptops. Usually we'll keep dev test runs small (multiples of 10) so they can iterate quickly, and CI large (multiples of 1000). So I'm not sure if this is the most common recommendation. Maybe it would be better to just acknowledge that there is a parameter to configure number of test cases and that it's good practice to configure this parameter separately between dev and CI environments based on the organizations needs?

proptest/src/test_runner/config.rs
```diff
@@ -350,7 +350,8 @@ pub struct Config {
     /// meaning of certain levels other than 0 is subject to change.
     ///
     /// - 0: No extra output.
-    /// - 1: Log test failure messages.
+    /// - 1: Log test failure messages. In state machine tests, this level is
+    ///   used to print transitions.
     /// - 2: Trace low-level details.
```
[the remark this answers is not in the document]
rex-remind101: Possibly add an issue for this after merging so we can revisit?

proptest-state-machine/src/strategy.rs
```diff
@@ -0,0 +1,761 @@
...
+            } else if self.transitions[ix].simplify() {
+                self.last_shrink = Some(self.shrink);
+                if self.check_acceptable(Some(ix)) {
+                    self.acceptable_transitions[ix] =
+                        (Current, self.transitions[ix].current());
+                    return true;
+                } else {
+                    let (state, _trans) =
+                        self.acceptable_transitions.get_mut(ix).unwrap();
+                    *state = SimplifyRejected;
```
rex-remind101, line 351: Why is it that we stop simplifying a transition as soon as we have a simplification that fails `check_acceptable`, should we not continue simplifying until we have an acceptable simplification which will later be `complicated()`?

The example from https://proptest-rs.github.io/proptest/proptest/tutorial/shrinking-basics.html continues to simplify on `// Still failing, find a simpler case` until the test succeeds or can no longer be simplified further.

rex-remind101: Reading this
```
/// Hence, the `preconditions` only needs to contain checks for invariants that
/// depend on the current state and may be broken by shrinking and it doesn't
/// need to cover invariants that do not depend on the current state.
```
adds some clarity, I suppose if the test can only ever operate within the transition graph described by the preconditions then there is no simplification allowed which breaks preconditions too

proptest-state-machine/src/strategy.rs
```diff
@@ -0,0 +1,761 @@
...
+                    matches!(state, SimplifyRejected | ComplicateRejected)
+                })
+    }
+
+    /// Find the next shrink transition. Loops back to the front of the list
+    /// when the end is reached.
+    fn next_shrink_transition(&self, current_ix: usize) -> Shrink {
+        if current_ix == self.max_ix {
+            // Either loop back to the start of the list...
+            Transition(0)
```
rex-remind101, line 470: Why is a rotation needed? After deleting from the end we start simplification at `self.shrink = Transition(0);` and step forward, do we not cover everything in one pass? Can we document why that is?

proptest-state-machine/src/strategy.rs
rex-remind101: Thinking through this it's still unclear to me. If a test fails because of an invalid transition given a state then wouldn't we'd want to simplify that transition as much as possible?

book/src/proptest/state-machine.md
```diff
@@ -0,0 +1,174 @@
...
+
+## How to use it
+
+Before using state machine testing, it is recommended to be at least familiar with the basic concepts of Proptest itself as it's built on its essential foundations. That is:
+
+- Strategies are composed and used to generate inputs to a state machine test.
+- A test will attempt to shrink the generated transitions sequence on a discovery of a case that breaks some properties.
+- It will capture regressions file with a seed that can be used to deterministically repeat the found case.
+
+In short, use `ReferenceStateMachine` and `StateMachineTest` to implement your state machine test and `prop_state_machine!` macro to run it.
```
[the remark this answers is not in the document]
tzemanovic: Thanks for pointing this out, though perhaps the issue is in the wording of the bullet points themselves. I'll try to make those clearer, because

> - Strategies are composed and used to generate inputs to a state machine test.

These are not state machine specific strategies - they are composed from regular proptest's foundations

> - A test will attempt to shrink the generated transitions sequence on a discovery of a case that breaks some properties.

The sequence of transitions is a strategy and like other strategies it does shrink to a minimal case that still reproduces the issue

proptest-state-machine/src/test_runner.rs
tzemanovic: good point, I'm looking forward to replacing the decl macros! :pray:

proptest/src/test_runner/runner.rs
tzemanovic: This is the only level that's being used directly in the state machine impl. It's kinda sub-optimal, but it's helpful for stepping through the state machine exec

proptest/src/test_runner/config.rs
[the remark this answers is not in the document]
tzemanovic: good call, opened https://github.com/proptest-rs/proptest/issues/323

proptest/src/test_runner/runner.rs
tzemanovic: there's a bit more context in this thread https://github.com/proptest-rs/proptest/pull/310#discussion_r1161370720

proptest-state-machine/src/strategy.rs
tzemanovic: If a shrink of a transition doesn't meet preconditions then we have to *omit* that change from the test as it's not a valid input. Sometimes a transition might become acceptable after a transition that comes before it in the sequence gets shrunk.

Together with the rotation on line 470, I experimentally found that this seems to shrink the transitions better than other strategies I tried before. The rotation ensures that we go back to transitions for which a shrink might have been previously rejected and it will attempt to simplify transitions as much as possible. I hope that answers those questions?

tzemanovic:
thanks for the additional reviews @cameron1024 and @rex-remind101! I've added couple fixups from the suggestions. Please let me know if you think this ready and I'll squash them

proptest-state-machine/src/strategy.rs
tzemanovic: I've added extra line here [bd6d95c](https://github.com/proptest-rs/proptest/pull/310/commits/bd6d95cc76c65b593da11d7135d057c64d90cb00), hope that's clearer

proptest-state-machine/src/strategy.rs
rex-remind101: I think I get what you're saying. Transitions form a causal chain of events, preconditions represent what paths of connected transitions can exist in that chain, and since causal chains are DAGs, our current set of transitions is one path in that DAG over the set of all _possible_ transitions (those which satisfy preconditions). Simplifying a transition may cause the test to enter a different path in the DAG overall so we have to do additional passes over the transition chain to test that it still meets preconditions.

I guess one thing that's still not entirely clear to me, are not failing preconditions a part of the test? If I have transitions A -> B -> C, and `B` fails preconditions from the start, then wouldn't I be encouraged to simplify `B` as much as possible until it succeeds and then complicate it (or just return it to its last failing state)? `B` is essentially a mediator of C, so observations on B screens off observations of C all else being equal, so we don't need to care about C at that point. I think what you may be saying is we simplify `A` first, and if that fails preconditions then we're screening off `B`, but `B` was what was failing the test to begin with and therefore we're no longer going to capture the original cause of test failure. Assuming I'm getting that right I think that makes sense, and possibly is a side-effect of the process we're using here to capture failure. Something we could possibly revisit in the future. 

(FYI, the reason I consider this a DAG is because i'm modeling this in my head via causality. E.g. Given a set of transitions InitialState -> A -> B -> D -> A (again), the causal notion should be InitialState -> A(t1) -> B(t1) -> D -> A(t2) -> B(t2) where observing the failure at B(t1) factors out everything that follows [inspired by [this](https://en.wikipedia.org/wiki/Causal_model)]).

rex-remind101 (approved):
I'd like to keep the conversation about how shrinking works going, but as-is I think this makes sense and looks very useful. Thanks for all the hard work you put in to this :)

proptest-state-machine/src/strategy.rs
tzemanovic: That's an interesting way to think about it! I think your causality model is on the right way, but any failing preconditions are in fact not part of the test (it is an invalid input for the test) and the initial transition sequence must always pass preconditions (this is ensured in `Sequential`'s `Strategy::new_tree` impl via filtering). What the shrinking process does is when it breaks a precondition is:

- for `Shrink::DeleteTransition` undoes deletion
- for `Shrink::Transition(_)` or `InitialState`  it ignores the simplification or complication and instead continues to use the state of the transition or of the initial state as it was before the shrink

Breaking a precondition on `Shrink::Transition` however doesn't stop the shrinking process and it will try to use that shrunk transition again next time the sequence rotation gets to it. In the example of `A -> B -> C`, if `B` shrink fails a precondition, `B` stays unchanged for that round. The shrinking process may get to it again after rotation and after `A` is shrunk it may (or may not) "unblock" the previously shrunk `B`'s strategy to become acceptable.

tzemanovic:
force-pushed to squash the fixups https://github.com/proptest-rs/proptest/compare/bd6d95cc76c65b593da11d7135d057c64d90cb00..d85ccf2bdc3a8813e1d7346880ddb4fbceddbe05

proptest-state-machine/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+homepage = "https://proptest-rs.github.io/proptest/proptest/state-machine.html"
+documentation = "https://docs.rs/proptest-state-machine/latest/proptest-state-machine/"
+keywords = ["property", "testing", "quickcheck", "fuzz", "state-machine"]
+categories = ["development-tools::testing"]
+description = """
+State machine based testing support for proptest.
+"""
+
+[dependencies]
+proptest = { version = "1.2.0", path = "../proptest" }
```
rex-remind101, line 17: I think CI may be busted because the latest crate version is 1.1.0 not 1.2.0

tzemanovic: I separated the needed proptest changes into https://github.com/proptest-rs/proptest/pull/324

russelldb:
Wow, this is so different from the original branch that I've been using for a year.

tzemanovic:
> Wow, this is so different from the original branch that I've been using for a year.

It’s most just renames (ref to the description) and being separated to its own crate

russelldb:
So far I've _failed_ to understand how to get it into my project…I use proptest = 1.2.0 in cargo toml (rather than my old fork) and everything breaks. I read the docs…no mention of how to include the new crate. I search on crates.io…can't find it. 

Is it "released" yet?

tzemanovic:
> So far I've _failed_ to understand how to get it into my project…I use proptest = 1.2.0 in cargo toml (rather than my old fork) and everything breaks. I read the docs…no mention of how to include the new crate. I search on crates.io…can't find it.
> 
> Is it "released" yet?

Yeah, it will be released as `v0.1.0`, but until then you can use it with `proptest-state-machine = { package = "proptest-state-machine", git = "https://github.com/proptest-rs/proptest.git", tag = "v0.1.2" }`

russelldb:
> Yeah, it will be released as `v0.1.0`, but until then you can use it with `proptest-state-machine = { package = "proptest-state-machine", git = "https://github.com/proptest-rs/proptest.git", tag = "v0.1.2" }`

Ah, so grateful for your help, many thanks! It will be great to get off this fork 😊

rex-remind101:
thank you for raising this @russelldb , it has been released https://crates.io/crates/proptest-state-machine

matheus23:
Thank you *SO* much for working on this @tzemanovic This is AMAZING!

PR #318, opened 2023-04-18
allow env vars to override non-default config

### LINKED ISSUE #315: allow to override explicit proptest_config from env vars
When the explicit `#[proptest_config(Config ...)]` is being used in a test, for any fields that are not `Config::default()` the `PROPTEST_` env vars are ignored. I think it would be better to allow the env vars to override the config as it enables ad hoc changes without recompiling, but it is potentially a breaking change if someone relies on the current behavior.

The fix should be easy to do using `fn contextualize_config` from `mod config`.

tzemanovic:
fixes #315

When using a non-default config or partially overriding the default config, the `proptest!` macro now allows to override config fields set from env vars, which gives the flexibility of being able to choose them without rebuilding.

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Breaking Changes
+
+- `PROPTEST_` environment variables now take precedence over tests' non-default
+  configuration.
+
 ### Bug Fixes
 
 ### New Features


matthew-russo (changes requested):
lgtm but lets update the Changelog as well

rex-remind101 (approved):
👍

matthew-russo (approved):

Issue #226, opened 2021-02-13
Why are u128/i128 unsupported in wasm32?


calebzulawski:
There's a single line in the book but nothing descriptive, and searching issues in this repo doesn't bring anything up either.  Thanks!

osa1:
I think it's just not implemented. I don't think there are any Wasm limitations that prevent us from implementing strategies for 128-bit types.

PR #322, opened 2023-05-15
[Fix] NonZero : only impl Arbitrary for 128-bit NonZero nums in non-wasm targets

### LINKED ISSUE #321: No longer builds for wasm as of v1.1
Build fails with the error below on v1.1. This regression was most likely introduced with 9c4647aae8ca9e4e67ef25e89a7ee08b858b608b.

```
error[E0277]: the trait bound `u128: arbitrary::traits::Arbitrary` is not satisfied
  --> /home/bruno/.cargo/registry/src/github.com-1ecc6299db9ec823/proptest-1.1.0/src/arbitrary/non_zero.rs:25:27
   |
25 |                 FilterMap<StrategyFor<$prim>, fn($prim) -> Option<Self>>;
   |                           ^^^^^^^^^^^^^^^^^^ the trait `arbitrary::traits::Arbitrary` is not implemented for `u128`
...
40 | non_zero_impl!(NonZeroU128, u128);
   | --------------------------------- in this macro invocation
   |
   = help: the following other types implement trait `arbitrary::traits::Arbitrary`:
             f32
             f64
             i16
             i32
             i64
             i8
             isize
             u16
           and 4 others
   = note: this error originates in the macro `non_zero_impl` (in Nightly builds, run with -Z macro-backtrace for more info)
```

matthew-russo:
fixes #321

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Bug Fixes
 
+- Don't implement Arbitrary for NonZeroU128 and NonZeroI128 on wasm targets where
+  u128 and i128 Arbitrary impls don't exist
+
 ### New Features
 
 ### Other Notes
--- a/proptest/src/arbitrary/_core/mod.rs
+++ b/proptest/src/arbitrary/_core/mod.rs
@@ -17,6 +17,7 @@ mod fmt;
 mod iter;
 mod marker;
 mod mem;
+mod non_zero;
 mod num;
 mod option;
 mod result;
--- a/proptest/src/arbitrary/_core/non_zero.rs
+++ b/proptest/src/arbitrary/_core/non_zero.rs
@@ -8,14 +8,15 @@
 // except according to those terms.
 
 use core::convert::TryFrom;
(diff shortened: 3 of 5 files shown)


rex-remind101 (approved):
👍

rex-remind101 (approved):
👍

cameron1024 (approved):
LGTM, thanks for this :grin: 

Out of curiosity, is there a reason why we don't implement `Arbitrary` for the 128-bit types on WASM? To be clear, I think it's the right call to revert the impls here (and I think it's also not a breaking change since it wasn't possible to rely on this impl anyways).

But maybe something worth thinking about for a future PR - I've worked with WASM a fair amount and I can't see a good reason why it couldn't support it :shrug:

matthew-russo:
> Out of curiosity, is there a reason why we don't implement Arbitrary for the 128-bit types on WASM? To be clear, I think it's the right call to revert the impls here (and I think it's also not a breaking change since it wasn't possible to rely on this impl anyways).

i don't think there's a good reason. we have #226 which has a comment that guesses that there isn't any technical blocker and it just needs to be implemented

PR #324, opened 2023-05-20
state-machine prerequisites

tzemanovic:
There are proptest crate-only changes separated from #310 that are needed for the new proptest-state-machine crate


matthew-russo (approved):
lgtm

rex-remind101 (approved):
great, thanks 👍

PR #325, opened 2023-05-22
bump version to 1.2.0

rex-remind101:
If this is good I will tag the commit with `v1.2.0`

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -19,6 +19,8 @@
 - Minimal failing input is now printed using debug pretty-printing
 - Made public `VarBitSet`, `SizeRange` read-only methods and num sampling
   functions in preparation for release of a `proptest-state-machine` crate.
+- Removed dependency on `quick_error`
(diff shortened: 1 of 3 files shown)


proptest/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+## 1.2.0
```
matthew-russo, line 3: did a quick scan over the commits since 1.1 to see if we're missing anything in changelog. seems like we dont have anything for:
- we're now publishing MSRV
- remove `quick_error` dependency

I can add a commit adding these shortly.

side note: just realized we had some changes to the derive crate but dont think we've published that yet or updated its changelog:
- updated `syn`, `quote` and `proc_macro2` to

matthew-russo: latest commit updates changelogs for both proptest and proptest-derive

matthew-russo (approved):

rex-remind101:
pushed the tag v1.2.0

PR #332, opened 2023-06-10
Arbitrary : negative impl of Arbitrary on never type (!)

### LINKED ISSUE #331: Handling Arbitrary on structs containing the Never type (!)
Hello! Here's a motivating example

```rust
fn main() {
    let x = any::<Wrapper<!>>();
}

#[derive(proptest_derive::Arbitrary, Debug)]
enum Wrapper<T> {
    A(T),
    B,
    C,
}
```

I'd like to be able to create such a `Wrapper` struct such that for `Wrapper<!>`, the proptest arbitrary picks between the other two options. This of course fails with
```
the trait `Arbitrary` is not implemented for `!`
```
Which makes sense! Impling arbitrary on `!` doesn't particularly make sense since it can't be instantiated. Perhaps I can implement `Arbitrary` manually with separate impls for `<T: Arbitrary>` and `T = !`.

```rust
type WrapperStrategy<T: Arbitrary> = impl Strategy<Value = Wrapper<T>>;
impl<T: Arbitrary> Arbitrary for Wrapper<T> {
    type Parameters = T::Parameters;
    type Strategy = WrapperStrategy<T>;

    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
        unimplemented!()
    }
}

impl Arbitrary for Wrapper<!> {
    type Parameters = ();
    type Strategy = WrapperStrategy<!>;

    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
        unimplemented!()
    }
}
```

Unfortunately, this also doesn't work because
```
19 | impl<T: Arbitrary> Arbitrary for Wrapper<T> {
   | ------------------------------------------- first implementation here
...
28 | impl Arbitrary for Wrapper<!> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `Wrapper<!>`
   |
   = note: upstream crates may add a new impl of trait `_IMPL_ARBITRARY_FOR_Wrapper::_proptest::arbitrary::Arbitrary` for type `!` in future versions
```

I briefly tried using https://doc.rust-lang.org/beta/unstable-book/language-features/negative-impls.html to `impl !Arbitrary for ! {}` - but this does not work for avoiding this coherence detection.

Any ideas on how this could work?
Perhaps `proptest_derive` could be augmented to detect this scenario and loosen the requirement that `T: Arbitrary`. Another idea could be something in the same vein as `#[proptest(skip)]` on `Wrapper::A(T)` that only skips it for the uninhabited case.

matthew-russo:
fixes #331

### DIFF
--- a/proptest/Cargo.toml
+++ b/proptest/Cargo.toml
@@ -56,6 +56,8 @@ atomic64bit = []
 # passed (see https://github.com/proptest-rs/proptest/issues/124).
 break-dead-code = []
 
+never-arbitrary = []
+
 [dependencies]
 bitflags = "1.0.1"
 unarray = "0.1.4"
--- a/proptest/src/arbitrary/primitives.rs
+++ b/proptest/src/arbitrary/primitives.rs
@@ -33,6 +33,9 @@ arbitrary!(f64, f64::Any; {
 
 arbitrary!(char, char::CharStrategy<'static>; char::any());
 
+#[cfg(all(feature = "unstable", feature = "never-arbitrary"))]
+impl !crate::arbitrary::Arbitrary for ! {}
+
 #[cfg(test)]
 mod test {
     no_panic_test!(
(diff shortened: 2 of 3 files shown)


proptest/Cargo.toml
```diff
@@ -56,6 +56,8 @@ atomic64bit = []
 # passed (see https://github.com/proptest-rs/proptest/issues/124).
 break-dead-code = []
 
+never-arbitrary = []
```
matthew-russo, line 59: Wasn't sure whether or not to just roll this in to the `unstable` feature flag. Opinions?

nipunn1313: Seems like a reasonable idea for stability as this is a fairly new area in the rust lang - (see my message below about coherence rules)

proptest/src/arbitrary/primitives.rs
```diff
@@ -44,3 +47,62 @@ mod test {
...
+        fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+            prop_oneof![
+                any::<T>().prop_map(Wrapper::A),
+                Just(Wrapper::B),
+                Just(Wrapper::C),
+            ].boxed()
+        }
+    }
+    
+    impl Arbitrary for Wrapper<!> {
```
nipunn1313, line 86: unfortunately, due to coherence rules, this won't be possible to implement if `Wrapper` is defined in another crate as the rust trait coherence rules considers it to be a conflicting implementation.

Without this PR the error looks like this
```
19 | impl<T: Arbitrary> Arbitrary for Wrapper<T> {
   | ------------------------------------------- first implementation here
...
28 | impl Arbitrary for Wrapper<!> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `Wrapper<!>`
   |
   = note: upstream crates may add a new impl of
trait `_IMPL_ARBITRARY_FOR_Wrapper::_proptest::arbitrary::Arbitrary` for type `!` in future versions
```

With this PR it still doesn't work, and the error looks like this
```
error[E0119]: conflicting implementations of trait `_IMPL_ARBITRARY_FOR_Wrapper::_proptest::arbitrary::Arbitrary` for type `Wrapper<!>`
  --> src/main.rs:13:10
   |
13 | #[derive(proptest_derive::Arbitrary, Clone, Debug)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `Wrapper<!>`
...
32 | impl Arbitrary for Wrapper<!> {
   | ----------------------------- first implementation here
   |
   = note: this error originates in the derive macro
`proptest_derive::Arbitrary`
(in Nightly builds, run with -Z macro-backtrace for more info)
```


There is an initiative inside of rustc to integrate negative impls into trait coherence rules, but until then this will be insufficient.
https://github.com/rust-lang/negative-impls-initiative/issues/1


I originally thought this would be a strict improvement, but actually judging from the error messages (IMO they end up regressing) - I think this is probably premature and should wait until the negative impls initiative furthers.

Perhaps hiding it behind another feature flag like you did is a good way to mitigate this concern!

nipunn1313 (approved):

PR #333, opened 2023-06-12
Remove `#![feature(core_intrinsics)]` usage

nvzqz:
This is for some reason enabled when using `alloc` instead of `std`, which prevents crates from using `proptest` on stable with only the `alloc` feature enabled.


tzemanovic:
thanks! It doesn't look like it's being used for anything, so I'm guessing this is historical left-over

tzemanovic (approved):

matthew-russo (approved):
lgtm

Issue #336, opened 2023-06-29
`bytes_regex` should permit generation of byte sequences that are invalid UTF-8


zackw:
This strategy function

```rust
fn invalid_ts() -> impl Strategy<Value = Vec<u8>> {
    prop::string::bytes_regex(
        r"(?s-u:|[^0-9].*|[0-9]+[^0-9.].*|[0-9]+\.[0-9]*[^0-9].*)"
    ).unwrap()
}
```

is _intended_ to generate, among other things, invalid UTF-8 byte sequences, because it's for testing a parser that works directly from data on disk that cannot be trusted.  What it actually _does_ is crash on the `unwrap()` with 

```none
thread 'attrs::test::parse_xattr_ts_invalid' panicked at 'called `Result::unwrap()` on an `Err` value:
    RegexSyntax(Translate(Error {
        kind: InvalidUtf8,
        pattern: "(?s-u:|[^0-9].*|[0-9]+[^0-9.].*|[0-9]+\\.[0-9]*[^0-9].*)",
        span: Span(Position(o: 7, l: 1, c: 8), Position(o: 13, l: 1, c: 14))
    }))'
```

Looking at the code, I _believe_ the change needed is for `bytes_regex` to have its own version of `regex_to_hir` that calls [`ParserBuilder::allow_invalid_utf8(true)`](https://docs.rs/regex-syntax/0.6.28/regex_syntax/struct.ParserBuilder.html#method.allow_invalid_utf8).

This change should have no visible effect on any existing code that uses `bytes_regex`, since one must also opt into generation of invalid UTF-8 within the regex itself (that's one of the things the `(?s-u:` does) and any existing regex that uses that flag must be using it in a way that actually _can't_ generate invalid UTF-8, or else they'd hit the same crash I'm hitting.

matthew-russo:
Hey sorry for the delay in getting to this. Gave this a look -- `bytes_regex` doesn't allow this *but* there is a public method `bytes_regex_parsed` that just takes in ` &regex_syntax::hir::Hir`, allowing you to generate that yourself with a customized `regex_syntax::ParserBuilder` which should address your needs. Given the workaround is straightforward and works generally for other regex parsing options, I'm inclined to leave things as is and not introduce a new, specific helper but I'm open to hearing arguments for that.

I'm going to close this but feel free to reopen if there are outstanding questions/discussions to be had

The proptest method:
[...]

PR #337, opened 2023-06-29
Permit use of (?-u) in byte-regex strategies (#336)

zackw:
It is desirable to be able to generate, from a regex, byte sequences that are not necessarily valid UTF-8.  For example, suppose you have a parser that accepts any string generated by the regex

```
[0-9]+(\.[0-9]*)?
```

Then, in your test suite, you might want to generate strings from the complementary regular language, which you could do with the regex

```
(?s:|[^0-9].*|[0-9]+[^0-9.].*|[0-9]+\.[0-9]*[^0-9].*)
```

However, this will still only generate valid UTF-8 strings.  Maybe you are parsing directly from byte sequences read from disk, in which case you want to test the parser’s ability to reject invalid UTF-8 _as well as_ valid UTF-8 but not within the accepted language.  Then you want this slight variation:

```
(?s-u:|[^0-9].*|[0-9]+[^0-9.].*|[0-9]+\.[0-9]*[^0-9].*)
```

But this regex will be rejected by `bytes_regex`, because by default `regex_syntax::Parser` errors out on any regex that potentially matches invalid UTF-8.  The application — i.e. proptest — must opt into use of such regexes.  This patch makes proptest do just that, for `bytes_regex` only.

There should be no change to the behavior of any existing test suite, because opting to allow use of `(?-u)` does not change the semantics of any regex that _doesn’t_ contain `(?-u)`, and any existing regex that _does_ contain `(?-u)` must be incapable of generating invalid UTF-8 for other reasons, or `regex_syntax::Parser` would be rejecting it. (For example, `(?-u:[a-z])` cannot generate invalid UTF-8.)

This patch also adds a bunch of tests for `bytes_regex`, which AFAICT was not being tested at all.  Some of these use the new functionality and others don’t.  There is quite a bit of code duplication in the test helper functions — `do_test` and `do_test_bytes` are almost identical, as are `generate_values_matching_regex` and `generate_byte_values_matching_regex`.  I am not good enough at generic metaprogramming in Rust to factor out the duplication.


cameron1024:
Thanks for the PR 😅 I'll have a look at this more closely when I have time, but upon first inspection it looks sensible.

If I haven't reviewed by the end of this weekend feel free to ping me 😅

tzemanovic (approved):
I think this is sound. Even though as @matthew-russo pointed out there is `bytes_regex_parsed` already, it requires to pull in `regex_syntax` to be used (side note, this crate should probably be re-exported as it appears in the public api), so extending the `fn bytes_regex` without a breaking change is a nice solution to allow this to be done directly. Additionally, having the extra doc on how to do this might help someone in future

matthew-russo:
Agh thanks for reminding me I need to look at this -- will do it this evening after work

matthew-russo (approved):

zackw:
Rebased on latest trunk, sorry for the delay.

zackw:
Fatfingered the close button, sorry for the noise.

PR #342, opened 2023-07-15
[Fix] Book : add empty main so mdbook doesn't wrap function in main

### LINKED ISSUE #251: Bug in published getting-started
From https://altsysrq.github.io/proptest-book/proptest/getting-started.html:

```
#![allow(unused)]
fn main() {
fn parse_date(s: &str) -> Option<(u32, u32, u32)> {
...
```

This actually does compile, though not after the test is added, since `parse_date` isn't in scope.

It looks like the version in [source control](https://github.com/AltSysrq/proptest/blob/master/book/src/proptest/getting-started.md) doesn't have the `main` wrapper. Maybe the published book needs to be rebuilt, or is being built from some other source?

matthew-russo:
fixes #251

### DIFF
--- a/book/src/proptest/getting-started.md
+++ b/book/src/proptest/getting-started.md
@@ -5,6 +5,8 @@ Let's say we want to make a function that parses dates of the form
 triple of integers is fine. So let's bang something out real quick.
 
 ```rust,no_run
+fn main() {}
+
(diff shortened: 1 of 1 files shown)


tzemanovic (approved):

rex-remind101 (approved):
👍

PR #352, opened 2023-07-29
tests: Use `allow` to suppress warnings.

waywardmonkeys:


tzemanovic:
let's fix the warnings instead of suppressing them #354

waywardmonkeys:
Agree that that is better.

PR #355, opened 2023-08-03
state-machine: fix logging by adding "std" feature

tzemanovic:
Before the feature was finalized, the state machine support was included in the core proptest crate that uses "std" feature. After crate split, the "std" feature was missing in the new crate so the code was dead.

The "std" feature is added to proptest-state-machine as a default feature allowing to switch the logging off in non-std env.


matthew-russo (approved):

rex-remind101 (approved):
👍

Issue #124, opened 2019-01-25
FYI: proptest breaks the compiler with RUSTFLAGS="-C link-dead-code"


boustrophedon:
This isn't a problem that proptest can fix (probably), but building proptest with RUSTFLAGS="-C link-dead-code" leads to an ICE. The link-dead-code flag is important for accurate code coverage results.

See rust-lang/rust#52478

AltSysrq:
0.9.1 now has a workaround. You need to opt out of the `break-dead-code` feature. If you're currently using the default feature set, you can do that with the following block in `Cargo.toml`:

```toml
[dependencies.proptest]
version = "0.9" # Only works for 0.9.1 or newer
default-features = false
# Enable all default features not known to break code coverage builds
features = ["default-code-coverage"]
```

I'm going to keep this issue open for now to help direct people to the workaround.

CAD97:
This has supposedly been fixed in https://github.com/rust-lang/rust/issues/58375, so if local testing shows the bug is fixed, this issue can be closed.

boustrophedon:
I can confirm that running `RUSTFLAGS="-C link-dead-code" cargo +1.37.0 build` fails and that it succeeds with all versions after 1.37. So this has actually been fixed since August :)

PR #357, opened 2023-08-03
proptest: remove no longer needed "break-dead-code" feature

tzemanovic:
clearing up a workaround from #124 that's no longer needed


rex-remind101 (approved):
nice 👍 though need to cleanup conflict please :)

matthew-russo (approved):

PR #360, opened 2023-08-17
Remove byteorder dependency

psychon:
The Rust standard library includes some byteorder-like helper functions for a while already. This commit replaces the use of byteorder with these helper functions.

----

I have no particular reason for this except for trying to remove a dependency. `byteorder` is not a "bad" dependency, but it still seemed unnecessary to me to do what is being done here.

Feel free to tell me that this is not worth it. :-)

### DIFF
--- a/proptest/Cargo.toml
+++ b/proptest/Cargo.toml
@@ -27,7 +27,7 @@ default-code-coverage = ["std", "fork", "timeout", "bit-set"]
 unstable = []
 
 # Enables the use of standard-library dependent features
-std = ["rand/std", "byteorder/std", "lazy_static",
+std = ["rand/std", "lazy_static",
        "regex-syntax", "num-traits/std"]
 
 # For use in no_std environments with access to an allocator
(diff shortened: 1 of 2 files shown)


matthew-russo (approved):
Changes look innocuous. In general I'm in favor of removing dependencies if the replacement is trivial logic.


For my own curiosity I want to go back and check commit history later to check the commit that added the dependency to see if there's any additional context.

psychon:
> to check the commit that added the dependency to see if there's any additional context.

`git log --reverse -p -S byteorder` says this was added in commit 4ba7873374f56f67e9c83f67a0135df4401a1bb1. That commit adds the dependency, but makes no use of it?!

Commit e9fe26ac7d568379e37cc55b49ec619a7052e59c adds it  as `extern crate`, but still makes no use of it.

The first real use appears in commit 6953a233a123f90f28df537ac3090db6605538b4 in a function `convert_to_new_format()`. It is used to convert `[u32; 4]` to `[u8; 16]` in little endian.

Dunno...

matthew-russo:
Ha that was quick -- thanks for checking.

cameron1024 (approved):
Thanks for this <3

PR #361, opened 2023-08-17
Make VarBitSet saturated + FromIterator efficient

sameer:
I noticed that `VarBitSet::saturated` takes a significant amount of time when generating data with `VecStrategy`. There are two perf issues:

1. not making use of the size hint in `FromIterator` leads to repeated grow calls
2. saturated should use `from_elem` since we're setting all bits

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - `regex-syntax` version 0.7 is now used.
 - Print a seed to stderr for a failed test even when a regressions file is already present.
+- Fixed a performance issue with `VarBitSet::saturated` that can slow down `VecStrategy`
 
 ## 1.2.0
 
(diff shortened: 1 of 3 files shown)


sameer:
I'm using [Instruments](https://help.apple.com/instruments/mac/10.0/#/dev7b09c84f5) on macOS to look at the CPU profile of a proptest and it's now optimized to `memset` 🚀 

Before:
Screenshot 2023-08-17 at 13 00 02

After:

Screenshot 2023-08-17 at 13 01 35

matthew-russo (approved):
Nice digging! This is a great find

sameer:
@matthew-russo are you able to approve the workflows / merge this?

matthew-russo:
I didn't realize the workflows weren't running. just kicked that off. we normally try to get two maintainers eyes on things before merging -- i pinged the others yesterday to give a look.

rex-remind101 (approved):
very nice 👍

Issue #272, opened 2022-10-25
impl Arbitrary for PathBuf


nipunn1313:
I see here - there's a TODO remaining.
I'd find it helpful in my project!
https://github.com/proptest-rs/proptest/blob/master/proptest/src/arbitrary/_std/path.rs

EricCrosson:
The `PathBuf` arbitrary was implemented in https://github.com/proptest-rs/proptest/pull/368 -- this issue can be closed 🎉

PR #362, opened 2023-08-21
Add PathBuf Arbitrary impl

maackle:
A minimal implementation of Arbitrary for PathBuf, per #272

### DIFF
--- a/proptest/src/arbitrary/_std/path.rs
+++ b/proptest/src/arbitrary/_std/path.rs
@@ -9,15 +9,22 @@
 
 //! Arbitrary implementations for `std::path`.
 
-use std::path::*;
+use std::{ffi::OsString, path::*};
+
+use crate::{arbitrary::StrategyFor, prelude::Strategy, strategy::MapInto};
 
 // TODO: Figure out PathBuf and then Box/Rc/Box<Path>.
 
 arbitrary!(StripPrefixError; Path::new("").strip_prefix("a").unwrap_err());
 
+arbitrary!(PathBuf, MapInto<StrategyFor<OsString>, Self>;
+    OsString::arbitrary().prop_map_into()
+);
+
 #[cfg(test)]
 mod test {
     no_panic_test!(
-        strip_prefix_error => StripPrefixError
+        strip_prefix_error => StripPrefixError,
+        path_buf => PathBuf
     );
 }


sunshowers:
(author of camino here, including the `Arbitrary` impl for `Utf8PathBuf`)

I think a useful impl should have a certain number of components, and should be able to generate relative and absolute paths with 50-50 probability. (And as a stretch goal it could also generate all of the other prefix components on Windows)

https://docs.rs/camino/1.1.6/src/camino/proptest_impls.rs.html#14-42

matthew-russo:
Sorry for the delay in getting to this -- works been busy this week. I'll try to give this a look tonight.

maackle:
Awesome @sunshowers, that's the kind of the feedback I was hoping for. "The best way to get something done right is to do it wrong on the internet". :)

proptest/src/arbitrary/_std/path.rs
```diff
@@ -9,15 +9,22 @@
 
 //! Arbitrary implementations for `std::path`.
 
-use std::path::*;
+use std::{ffi::OsString, path::*};
+
+use crate::{arbitrary::StrategyFor, prelude::Strategy, strategy::MapInto};
 
 // TODO: Figure out PathBuf and then Box/Rc/Box<Path>.
```
matthew-russo, line 16: we should edit/remove this comment once we get to a spot we're happy with

proptest/src/arbitrary/_std/path.rs
```diff
@@ -9,15 +9,22 @@
...
 
 arbitrary!(StripPrefixError; Path::new("").strip_prefix("a").unwrap_err());
 
+arbitrary!(PathBuf, MapInto<StrategyFor<OsString>, Self>;
+    OsString::arbitrary().prop_map_into()
+);
```
matthew-russo, line 22: agreed with @sunshowers comment that we should be a bit more robust in the impl here. I agree that a good starting place is to ensure we get a spread of path segments and we get coverage of absolute and relative paths. I'd be fine with merging a strategy that hits those two marks. I think more robust edge case generation could be done separately.

let me know if you'd be willing to tackle those in a follow up

matthew-russo (changes requested):

proptest/src/arbitrary/_std/path.rs
maackle: Unfortunately I won't have bandwidth to take this on probably for quite a while, but I hope this can be a very minimal bit of inspiration for others to build upon though.

Also I might point out that even though this is not an ideal implementation in terms of covering a wide range of possibilities, it *does* produce an arbitrary PathBuf, so maybe in the spirit of "something is better than nothing", this could be merged with the intention to improve upon it. Maybe even with a deprecation to make it clear that this is not a good implementation, but is currently the best we have? That would get me unstuck on my own project, where I really just need a completely arbitrary PathBuf, and have some types that contain PathBufs.

sunshowers: Could either of you copy over the impl in camino? Otherwise I can try getting a PR up with that impl when I have time.

sunshowers: Ended up doing this myself, see https://github.com/proptest-rs/proptest/pull/368.

matthew-russo: Thanks! I'm on vacation until next week so may not get a chance to look at it but will try to look sometime next week if I don't find any time before then

PR #365, opened 2023-08-24
[proptest-derive] disable clippy::arc_with_non_send_sync

sunshowers:
This is a new lint added in Rust 1.72. See comment for why.


matthew-russo (approved):

cameron1024 (approved):
Thanks :)

PR #367, opened 2023-08-31
Book tips and best practices

hamiltop:
Also update the book template and config. Lots of changes here, very open to feedback.


book/src/proptest-derive/modifiers.md
```diff
@@ -34,20 +34,31 @@ of [`prop_filter`].
 Example:
 
 ```rust
+# extern crate proptest_derive;
+# extern crate proptest;
+# use proptest_derive::Arbitrary;
+# use proptest::prelude::*;
```
rex-remind101, line 40: what do all these pound signs do?

book/src/proptest/tips-and-best-practices.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Reusing mutable resources 
+Sometimes you may want to reuse mutable resources across individual cases. For example, you may want
+to reuse a database connection or a file handle to avoid the overhead of opening and closing it for
+each case. Because the `proptest!` macro (when used with closure-style invocation) requires a `Fn`, you need to wrap your state in a `RefCell`:
+
+```rust,mdbook-runnable
+use std::cell::RefCell;
+
+# struct MyState {};
```
rex-remind101, line 26: `MyState` isn't actually used in the test it looks like. In terms of communicating what's happening, i'm not sure the `My` prefix is that helpful either and I'm guessing if `State` has `connect` it's really `ConnectionPool`? what do you think about using that name?

book/src/proptest/tutorial/compound-strategies.md
```diff
@@ -18,6 +19,11 @@ fn test_add() {
     let mut runner = TestRunner::default();
     runner.run(/* uhhm... */).unwrap();
 }
+# #[cfg(not(test))]
+# fn test_add() {
+#     let mut runner = TestRunner::default();
+#     runner.run(/* uhhm... */).unwrap();
+# }
```
rex-remind101, line 26: what's does adding a not test version of the same code do?

book/src/proptest-derive/modifiers.md
rex-remind101: I take it from reading this later `// doctests don't run under cfg(test), so we need to do this.` that this hides this code in the docs but it's useful for compilation of the doc test?

book/src/proptest/tutorial/compound-strategies.md
rex-remind101: so does this compile and run in the background make it appear to the user that the `#[test]` version ran where actually this did?

rex-remind101 (commented):
just a few questions out of curiosity, this looks quite good to me

book/src/proptest-derive/modifiers.md
hamiltop: Yeah. This is a weird hack,  but you got the main idea.

Pound signs hide these lines from the rendered docs, so we just show the bit that's useful for people to see.

And because doctests aren't compiled under cfg test, the test cases don't get defined and can't be called from main unless we also define them with cfg not test.

book/src/proptest/tips-and-best-practices.md
hamiltop: Looking at this now, I think this example is broken. I'll revisit.

book/src/proptest/tutorial/compound-strategies.md
hamiltop: Yeah, the actual doctest is the cfg not test fn. The test function is just for display purposes. There's probably a better way to DRY this up. I'll take another pass at it.

hamiltop:
@rex-remind101 Took a second pass through and found a less hacky-hack to deal with the cfg test stuff. I also enabled a bunch more proptests.

One additional quirk: `mdbook test` cannot naively handle extern crates, so to run the tests you have to run it with `mdbook test -L ../target/debug/deps`

It might be worth exploring https://github.com/tfpk/mdbook-keeper/ (which solves many of these quirks).

rex-remind101 (approved):
looks good 👍

hamiltop:
Who can approve the workflow so CI can run?

matthew-russo (approved):
Thanks for cleaning this up and the new additions!

PR #368, opened 2023-09-01
Add PathBuf Arbitrary impl with tests

sunshowers:
Followup to #362. Includes documentation. I'm pretty new to this and I had to cargo-cult a bit so please point out issues!

Thanks to Michael for starting work on this!

### DIFF
--- a/proptest/src/arbitrary/_std/path.rs
+++ b/proptest/src/arbitrary/_std/path.rs
@@ -9,15 +9,72 @@
 
 //! Arbitrary implementations for `std::path`.
 
-use std::path::*;
+use std::{ffi::OsString, path::*};
 
-// TODO: Figure out PathBuf and then Box/Rc/Box<Path>.
+use crate::{
+    arbitrary::{SMapped, StrategyFor},
+    prelude::{any, any_with, Arbitrary, Strategy},
+    std_facade::{string::ToString, Arc, Box, Rc, Vec},
+    strategy::{statics::static_map, MapInto},
+};
 
 arbitrary!(StripPrefixError; Path::new("").strip_prefix("a").unwrap_err());
+
+/// This implementation takes two parameters: a range for the number of components, and a regex to
+/// determine the string. It generates either a relative or an absolute path with equal probability.
+///
+/// Currently, this implementation does not generate:
+///
+/// * Paths that are not valid UTF-8 (this is unlikely to change)
+/// * Paths with a [`PrefixComponent`](std::path::PrefixComponent) on Windows, e.g. `C:\` (this may
+///   change in the future)
+impl Arbitrary for PathBuf {
+    type Parameters = <Vec<OsString> as Arbitrary>::Parameters;
+    type Strategy = SMapped<(bool, Vec<OsString>), Self>;
+
+    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+        static_map(
+            (any::<bool>(), any_with::<Vec<OsString>>(args)),
+            |(is_absolute, parts)| {
+                let mut out = PathBuf::new();
+                if is_absolute {
+                    out.push(&MAIN_SEPARATOR.to_string());
+                }
+                for part in parts {
+                    out.push(&part);
+                }
+                out
+            },
+        )
+    }
+}
+
+macro_rules! dst_wrapped {
+    ($($w: ident),*) => {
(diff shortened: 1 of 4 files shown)


sunshowers:
For me a number of tests failed, I'm not sure why: https://gist.github.com/sunshowers/5ad72397fe70205bbb27cae534d65443#file-gistfile0-txt-L4551-L4685

But those tests also failed for me on main.

(This is with `cargo nextest run --no-fail-fast`)

sunshowers:
I have a feeling this isn't quite correct. I think it might generate paths that are too long.

proptest/src/arbitrary/_std/path.rs
```diff
@@ -9,15 +9,72 @@
...
+/// * Paths that are not valid UTF-8 (this is unlikely to change)
+/// * Paths with a [`PrefixComponent`](std::path::PrefixComponent) on Windows, e.g. `C:\` (this may
+///   change in the future)
+impl Arbitrary for PathBuf {
+    type Parameters = <Vec<OsString> as Arbitrary>::Parameters;
+    type Strategy = SMapped<(bool, Vec<OsString>), Self>;
+
+    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+        static_map(
+            (any::<bool>(), any_with::<Vec<OsString>>(args)),
```
matthew-russo, line 37: i'm not familiar with the `OsString` arbitrary impl but I'm wondering if it can generate a string with a delimiter in it. will pull this locally and mess around more. otherwise this looks good.

matthew-russo: OsString strategy does indeed generate strings with the delimiter.

sunshowers: I think that's okay, the number of delimiters is not a hard and fast rule here.

I'm more concerned about the number of components being too large by default.

I'm wondering if we should define a PathParameters struct with a custom default value for the number of delimiters.

matthew-russo: I did notice that quite a few components were getting generated. I think a `PathParameters` struct with a default number and ensuring the output has that number (which may involve trimming any excess generated) could make sense.  I think it'd be an un-intuitive if a user asks for `n` segments and we give them `n` + `m`.

sunshowers: That makes sense. You're welcome to work on it, or I'll get around to it at some point later this week (speaking at RustConf in 3 days 😅)

sunshowers: @matthew-russo OK, done. Let me know if this makes sense.

proptest/src/arbitrary/_std/path.rs
```diff
@@ -11,13 +11,121 @@
...
+    arbitrary::{SMapped, StrategyFor},
+    path::PathParams,
+    prelude::{any, any_with, Arbitrary, Strategy},
+    std_facade::{string::ToString, Arc, Box, Rc, String, Vec},
+    strategy::{statics::static_map, MapInto},
+};
 
 arbitrary!(StripPrefixError; Path::new("").strip_prefix("a").unwrap_err());
 
+/// A private type (not actually pub) representing the output of [`PathParams`] that can't be
```
matthew-russo, line 24: non-blocking: it is actually pub(?) not sure if this just means it isn't intended to be publicly used or if this is residual from previous iterations.

matthew-russo (approved):
lgtm. thanks for the contribution!

proptest/src/arbitrary/_std/path.rs
matthew-russo: Sorry work and personal life have kept me quite busy so I missed your last message. Thanks for getting back to this and congrats on the talk!

proptest/src/arbitrary/_std/path.rs
sunshowers: I meant that even though it is pub, it is not exported as an API so that backwards compatibility is maintained. Any suggestions for better phrasing?

matthew-russo: I don't think its a huge deal either way. I would probably just say something like:
`"An opaque type representing the output of ...`

rex-remind101 (approved):
this all looks good to me 👍

Issue #82, opened 2018-08-18
Common problematic values are often unlikely to be generated


shelbyd:
In the Readme, the "Limitations of Property Testing" section details that the `any::<i64>()` strategy is unlikely to produce `i64::MIN`. This, and other i64 constants are usually far more likely to cause problems than an arbitrary i64.

I propose, at least for the built in Strategies, that they proactively generate common problematic values before generating arbitrary ones.

Some possibilities for the different types:

- **integer**: 0, -1, 1, MAX, MIN, MAX - 1, MIN + 1
- **range**: start, end, start + 1, start - 1
- **float**: 0, -0, INFINITY, NEG_INFINITY, MAX, MIN, MIN_POSITIVE, NAN
- **string**: "", common control characters, rtl unicode, others from the [Big List of Naughty Strings](https://github.com/minimaxir/big-list-of-naughty-strings)
- **slice**: fewest elements, most elements

I'm happy to work on this. I'll just need a bit of direction in how to go about it.

AltSysrq:
I'd certainly welcome any contributions!

> integer, range

These all sound reasonable to me. Integers are defined in terms of ranges so instead of `MIN` and `MAX`, the range boundaries are probably the best option. This would just be a matter of changing how the [starting value is generated](https://github.com/AltSysrq/proptest/blob/db192a4558600d2b8cd73ba241902cfaaeb21764/src/num.rs#L85) (`ANY` is [special](https://github.com/AltSysrq/proptest/blob/db192a4558600d2b8cd73ba241902cfaaeb21764/src/num.rs#L46) but might not need to be anymore).

> float

Some of these already happen (zeroes, infinites, NaNs), though `any::<$float>()` by design [only includes real numbers](https://github.com/AltSysrq/proptest/blob/db192a4558600d2b8cd73ba241902cfaaeb21764/src/arbitrary/primitives.rs#L26) since (in my experience at least) the majority of code handling floats isn't expected to do anything reasonable if handed infinity or NaN. Floats are otherwise handled similarly to integers and only the [starting value](https://github.com/AltSysrq/proptest/blob/db192a4558600d2b8cd73ba241902cfaaeb21764/src/num.rs#L562) would need to be changed.

> slice

Similar to integer range handling, [all that needs to change is the starting value](chttps://github.com/AltSysrq/proptest/blob/db192a4558600d2b8cd73ba241902cfaaeb21764/src/num.rs#L562)

> string ""

[...]

PR #369, opened 2023-09-06
Add edge bias feature to bias for testing edge cases.

ajantti:
It works by randomly changing some generated values to be one of the random edge cases. The first tests are done with 100% edge cases, and then the chance goes down over the number of test cases until it's 0%.

The average amount of edge case tests can be controlled by edge_bias configuration option. It defaults to 0.25. (25% of generated values are edge cases over all the tests.)

In failure the edge bias is persisted to the persistence file along with the seed, so it should be reproducible.

So far it only does ints, floats and ranges.

I also got some validation for the system, because some of the existing tests started failing.

For example the time::Duration test failed because if both secs and nanos are MAX, the nanos will overflow into secs, and then secs will overflow and the function panics.

Note that I'm not sure if the fix I made is the best, since I just mod by a billion, so it biases the random distribution a bit. I don't know if this is undesireable.

float_samplers test subbsequent_splits_always_match_bounds also started failing (for f32 and f64 versions both) because it couldn't generate multiple intervals for values like
```
minimal failing input: (low, high) = (
    1.7976931348623155e308,
    1.7976931348623157e308,
), index = Index(
    0,
)
```
I fixed it by checking for intervals.count == 1, but I feel like that's not the best solution. From reading the other tests, it feels like the intervals are supposed to be able to be split, but in the end I didn't look fully into it.

I also think the tests should include nans and infinities as edge cases, but I didn't add those because it would cause a lot of things to fail, because they're made with the assumption that those are not generated. (Or expecting and asserting.)

Except for the strings, this should implement https://github.com/proptest-rs/proptest/issues/82

I just started with Rust, so I hope there's nothing too stupid in there...

### DIFF
--- a/proptest/src/arbitrary/_std/time.rs
+++ b/proptest/src/arbitrary/_std/time.rs
@@ -17,8 +17,11 @@ use crate::num;
 use crate::strategy::statics::{self, static_map};
 
 arbitrary!(Duration, SMapped<(u64, u32), Self>;
-    static_map(any::<(u64, u32)>(), |(a, b)| Duration::new(a, b))
-);
+    static_map(any::<(u64, u32)>(), |(a, b)|
+    // Duration::new panics if nanoseconds are over one billion (one second)
+    // and overflowing into the seconds would overflow the second counter.
+    Duration::new(a, b % 1_000_000_000)
+));
 
 // Instant::now() "never" returns the same Instant, so no shrinking may occur!
 arbitrary!(Instant; Self::now());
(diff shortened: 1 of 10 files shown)


matthew-russo:

NB: #504 was just recently opened with a similar topic

0rzech:
Would it be possible to add strategy transformation which allows specifying edge cases manually, like [jqwik does](https://jqwik.net/docs/current/user-guide.html#configuring-edge-cases-themselves)?

PR #376, opened 2023-10-05
Follow new lint rule in macro generation

hamiltop:
clippy::ignored_unit_patterns is a new lint rule that the generated code seems to be violating. This fixes it so the generated code passes lint checks that the end user wants.


rex-remind101 (approved):
👍

matthew-russo (approved):
lgtm

Issue #379, opened 2023-10-12
Update regex-syntax to 0.8


JustusFluegel:
Update regex-syntax to 0.8 and regex to the corresponding version. This would be nice as proptest is the only crate that still requires regex-syntax v0.7 in some of my projects and I would appreciate not having to compile it in that version anymore.

I would also be open to try to create a pr but I am not very familiar with the proptest codebase  so I don't know how that would go. Feel free to tell me if that is welcome tho.

matthew-russo:
Addressed with:
https://github.com/proptest-rs/proptest/pull/386

PR #381, opened 2023-10-12
Depend on regex-syntax 0.7 or 0.8

sjackman:
Depend on `regex-syntax` 0.7 or 0.8.

### DIFF
--- a/proptest/Cargo.toml
+++ b/proptest/Cargo.toml
@@ -75,7 +75,7 @@ features = ["libm"]
 [dependencies.regex-syntax]
 # If you change this, make sure to also bump the `regex` dependency to a
 # version that also uses this version of regex-syntax.
-version = "0.7"
+version = ">=0.7, <0.9"
 optional = true
 
 [dependencies.bit-set]


matthew-russo (approved):

rex-remind101 (approved):
👍  once https://github.com/proptest-rs/proptest/pull/382 is in and ci passes

matthew-russo:
If you get a chance, can you rebase and update your branch so we can get CI running? I tried to do it but i can't push to your branch

matthew-russo:
and fixes #379

matthew-russo:
this actually got addressed by our recent merge of dependabot:

https://github.com/proptest-rs/proptest/pull/386

sjackman:
> If you get a chance, can you rebase and update your branch so we can get CI running? I tried to do it but i can't push to your branch

This PR is different than @dependabot's PR #386 in that this PR permits either `regex-syntax` version 0.7 or 0.8, whereas @dependabot's PR permits only version 0.8. I've rebased this PR onto `origin/master`. Would you please reopen this PR and merge it? 
- https://github.com/proptest-rs/proptest/pull/386

sjackman:
> Is there a use case for still requiring 0.7? I expected a dependency on 0.8 to suffice.

A more permissive dependency allows a project to use either `regex-syntax` version 0.7 or 0.8, so that if that project hasn't yet updated their direct dependency on `regex-syntax` from version 0.7 to 0.8, they don't need to install two versions of `regex-syntax`, one for the direct dependency of their project, and one for the indirect dependency through `proptest`.

Dependabot supports updating `version = ">=0.7, <0.9"` to `version = ">=0.7, <0.10" in a Dependabot PR.

PR #380, opened 2023-10-12
Enable Dependabot

sjackman:
Enable [Dependabot](https://docs.github.com/en/code-security/dependabot) to open PRs to update dependencies.

PR #377, opened 2023-10-09
Fix prop_state_machine cannot find value `config`

sameer:
When `#![proptest_config(..)]` is not provided, the macro should use the default proptest config. This PR fixes an issue that prevents this from working -- there's a reference to `config` that should be `Default::default()` instead:
```
error[E0425]: cannot find value `config` in this scope
     |
2132 | /     prop_state_machine! {
2133 | |         #[test]
2134 | |         fn run(sequential 1..20 => MyTest);
2135 | |     }
     | |_____^ not found in this scope
     |
     = note: this error originates in the macro `prop_state_machine` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #382, opened 2023-10-14
Build : updates MSRV to 1.65 and update CI to use rust 1.65

matthew-russo:
unblocks #380 and #381


matthew-russo:
also unblocks #377

rex-remind101 (approved):
👍

Issue #387, opened 2023-10-17
Random Shrinking in state machine tests


henriiik:
Hello!

I am using `proptest_state_machine`  to write some tests. Whilst experimenting, i noticed that it appears that the input for shrinking the collection of transitions is the original collection of transitions, rather than the transitions that were able to be executed against the system under test.

here is a portion of the verbose log output:

```
proptest: Next test input: (initial_state, transitions) = (RefState { wallet_state: None }, [RegisterWallet, RegisterAlreadyRegisteredWallet, DeregisterWallet, RegisterDeregisteredWallet, DeregisterWallet, RegisterDeregisteredWallet, DeregisterWallet, RegisterDeregisteredWallet])

Running a test case with 8 transitions.

Applying transition 1/8: RegisterWallet

Applying transition 2/8: RegisterAlreadyRegisteredWallet
thread 'experiment::test_fn' panicked at libs/spbt/src/experiment.rs:178:17:
OMG2!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
proptest: Test case failed: OMG2!
proptest: Next test input: (initial_state, transitions) = (RefState { wallet_state: None }, [RegisterWallet, RegisterAlreadyRegisteredWallet, DeregisterWallet, RegisterDeregisteredWallet, DeregisterWallet, RegisterDeregisteredWallet, DeregisterWallet])

Running a test case with 7 transitions.

Applying transition 1/7: RegisterWallet

Applying transition 2/7: RegisterAlreadyRegisteredWallet
thread 'experiment::test_fn' panicked at libs/spbt/src/experiment.rs:178:17:
OMG2!
proptest: Test case failed: OMG2!
```

As you can see after the abstract execution proptest generated a sequence of 8 transitions. When executing those transitions against the system under test it failed after the second transition. proptests current behavior appears to be attempting to shrink the original sequence of transitions from the abstract execution (8 transitions) rather than the transitions that were executed against system until it failed (2 transitions).

Is this behavior deliberate? Would it not be better to shrink based on the sequence of transitions that led to failure as this is more likely to lead to a minimally reproducible case?

Additionally it appears from my understanding of the code, that it only shrinks from the back until it removes the bad transition, and does not try to remove random transitions from before it.

https://github.com/proptest-rs/proptest/blob/d6f95d46e7d1a1875d16b6f9eb91906b52a5b8c6/proptest-state-machine/src/strategy.rs#L295-L322

[...]

PR #388, opened 2023-10-19
Delete unseen transitions in state machine test

henriiik:
This PR adds a new shrinking step to `proptest_state_machine::SequentialValueTree` that deletes all transitions not seen by the test runner as the first step. This is related to #387.

To do that it changes the value type from a `Vec<Transition>` to a new wrapper type `ObservedVec<T>`. This new type contains a counter that is shared with `SequentialValueTree` and keeps track of how many transitions have been seen by the test runner.

The shared counter is reset on each call to `simplify` / `complicate`, therefore it is no longer valid to call `current` multiple times without calling `simplify` and/or `complicate` in between.

Given my (admittedly limited) understanding of how `proptest` works this is not an issue as the intention is for the value tree to be used in a loop where each iteration starts with a call to `current` which produces a value that is used in a test run and based on the test result `simplify` or `complicate` is called at least one time before the next iteration.

Any feedback would be welcome!

### DIFF
--- a/proptest-state-machine/src/strategy.rs
+++ b/proptest-state-machine/src/strategy.rs
@@ -9,6 +9,9 @@
 
 //! Strategies used for abstract state machine testing.
 
+use std::sync::atomic::{self, AtomicUsize};
+use std::sync::Arc;
+
 use proptest::bits::{BitSetLike, VarBitSet};
 use proptest::collection::SizeRange;
 use proptest::num::sample_uniform_incl;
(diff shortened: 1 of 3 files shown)


tzemanovic:
thank you @henriiik, will take a look

proptest-state-machine/src/strategy.rs
```diff
@@ -575,7 +671,7 @@ mod test {
     ///
     /// This constant can be determined from the test
     /// `number_of_sequential_value_tree_simplifications`.
-    const SIMPLIFICATIONS: usize = 699;
+    const SIMPLIFICATIONS: usize = 700;
```
rexmas, line 674: this change was recently merged, so we'll need to resolve these two https://github.com/proptest-rs/proptest/pull/401/files#diff-4215d2d96283ff524e989ed16ad7059ed80dfc8e0c7c8d0467f549700fe7fdf2R570

rexmas (approved):
left one comment, the rest of the changes make a lot of sense to me. @tzemanovic knows this code best though so we'll wait for his feedback :)

thanks for the contribution 👍

proptest-state-machine/src/strategy.rs
tzemanovic: Hey, thanks for your patience. There's another thing that's changed in the last release that affects this PR - the `min_size` has been removed from `SequentialValueTree`. The changes in this PR are using it in `fn try_simplify` and `test_value_tree_initial_simplification`. I've started to rebase this PR, but I also have some other comments that I'd like to address

tzemanovic:
I've rebased the branch on `proptest-state-machine-0.2.0` and added couple fix-ups ([4c8696d](https://github.com/proptest-rs/proptest/pull/388/commits/4c8696d9e470b6159711108ae43c616d6886c2fa) and [0aa8087](https://github.com/proptest-rs/proptest/pull/388/commits/0aa8087331c071d3f58506086941759d83a416b6)).

I've got some more changes that I'd like to add WIP, will update here soon

proptest-state-machine/src/strategy.rs
tzemanovic: I addressed this after rebase in https://github.com/proptest-rs/proptest/pull/388/commits/4c8696d9e470b6159711108ae43c616d6886c2fa

henriiik:
Thanks for picking this up, let me know if you want me to help in any way :)

tzemanovic:
Thanks again @henriiik, I'd like to add the following changes:

 - Return the ref to `seen_transitions_counter` from `fn current` separately without wrapping the transitions in `ObservedVec` and increment the counter from `StateMachineTest::test_sequential`.
 - Because we can assume that the test results are deterministic (otherwise the shrinking process is not useful), I think we can safely remove the unseen transitions without undoing it in `fn complicate`. This is unlike the other `Shrink` cases where we don't know if the shrinking will affect the result. Furthermore, we only need to count the seen transitions before shrinking as it shouldn't have any effect anymore when we're removing transitions 1-by-1.
 - Remove the first panicking check in `fn current` - it prevents multiple refs to `seen_transitions_counter` which should not be necessary, we can allow multiple refs if needed for customization. I kept the second condition, but slightly changed the message.
 
 The last 2 fixup commits make these changes, if you could pls check them over to make sure that this still works for your use-case.

tzemanovic:
I've opened #434 with the fixups squashed, I can push it back here if you're happy with the changes

henriiik:
@tzemanovic  Looks good to me!

tzemanovic (approved):

PR #401, opened 2023-11-22
state-machine: prep for a minor release

tzemanovic:


proptest-state-machine/src/strategy.rs
```diff
@@ -567,7 +567,7 @@ mod test {
     ///
     /// This constant can be determined from the test
     /// `number_of_sequential_value_tree_simplifications`.
-    const SIMPLIFICATIONS: usize = 699;
+    const SIMPLIFICATIONS: usize = 32;
```
rexmas, line 570: for clarity why this change?

tzemanovic: I missed this change in #392 that fixed a bug in shrinking, it was actually failing on that branch

rexmas (approved):
👍

Issue #152, opened 2019-06-24
Tuple unions created by proptest_derive should branch out more (stack overflow while generating values)


sunshowers:
Hi again! ☺️

So it turns out that we have an enum with 70+ different variants, and it's large enough that generating values for it appears to cause a stack overflow with the default Rust stack size 😕

At the moment, when there are a large number of variants, proptest_derive will create essentially a linear structure. For variants `v<n>`, the structure looks like:

```
[v0, v1, v2, v3, ... v8, ___ ]
                          |
                          v
                         [v9, v10, v11, ...]
```

and so on. Instead, what we could do is a bit more branchy:

```
[ ______________ , _______________ , ...]
       |                  |
       v                  v
[v0, v1, v2, ...] [v10, v11, v12, ...]
```

There are several possible distributions we could use, e.g. an exponential distribution (1 variant in the first slot, 2 in the second slot and so on). I'd love to hear your opinions on it! Thank you 😀

This isn't a problem with `prop_oneof!` since it just uses the `Union` strategy past 10 variants.

sunshowers:
I might give implementing this a shot if I have free time soon.

Centril:
(Might also be relevant to https://github.com/AltSysrq/proptest/pull/151, cc @alessandrod)

Well... this was unexpected (for me) =P
I'm curious to hear more about why (and where) the stack overflow happens.

> Instead, what we could do is a bit more branchy:

Did you try the more branchy expansion in a manual implementation to see that the issue is resolved with that?

> There are several possible distributions we could use, e.g. an exponential distribution (1 variant in the first slot, 2 in the second slot and so on). I'd love to hear your opinions on it! Thank you 😀

Here's two alternatives that come to mind (the invariant here is that we have max 3 slots):

```
[ ____________, v4]
       |
       v
  [v0, v1, v2]

[ ____________, ____________]
       |             |
       v             v
  [v0, v1, v2]   [v4, v5]
```
(this seems most even)

vs.
```
[ ________, ________]
     |         |
     v         v
  [v0, v1]   [v3, v4]

[ ____________, v4, v5]
       |
       v
  [v0, v1, v2]
```
(this seems simplest; tho you might want to mirror the to put "extras" to the right instead...)

matthew-russo:
an interesting note. I had a pretty old toolchain (1.72 nightly) and could repro this. updating to 1.77 nightly turned the stack overflow in to a segfault. then if i removed a few variants from my test then it turned back in to a stack overflow.

Issue #249, opened 2021-08-11
ValueTree lives entirely on stack (stack overflow situations)


rbtcollins:
#152 talks about a specific case of stack overflow generating the ValueTree, but it seems to me that the crux of the issue is the reliance on a stack-only datastructure for ValueTree. At least, presuming I've read the code correctly :)  If the built up valuetree was on-heap, frames would require much less footprint even with unbalanced structures - and the result should ideally be able to reside either on-stack or on-heap.

filing this because I have a gnarly big datastructure (openapi specifications) that is mutually recursive along several paths and while any fully realised doc should be under 2MB, it is easy to imagine the bookkeeping of valuetree + a recursed datastructure exceeding that easily. Note that my case passes with cargo test --release and a default stack size; strongly suggesting copy optimisations reducing the footprint are a factor.

aalexandrov:
We've been hit by the same issue in https://github.com/MaterializeInc/materialize.

aalexandrov:
FYI, the `MaterializeInc/proptest` fork now contains [a dedicated commit](https://github.com/proptest-rs/proptest/commit/fc9660f0f45ad49949d42341f964597a44e5fce0) that resolves the issue by adding a `boxed_union` feature, which when enabled will change the `enum` expansion in `#[derive(Arbitrary)]` to use `Union<BoxedStrategy<Self>>` instead of `TupleUnion`.

If enough people show interest, I will open a PR against the main repository with that.

rex-remind101:
This commit seems like a relatively straight forward addition, I say open a pr. cc @matthew-russo

aalexandrov:
> This commit seems like a relatively straight forward addition, I say open a pr, thanks! cc @matthew-russo

Sorry for dropping the ball on this. I'll try to get the PR in before the end of the week.

rexmas:
follow up -- we opened a pr with your changes and included you as the author. thanks for the hard work you put in! https://github.com/proptest-rs/proptest/pull/406

aalexandrov:
> follow up @aalexandrov -- we opened a pr with your changes and included you as the author. thanks for the hard work you put in! #406

I realized that when I tried to apply my commits on the upstream `master` branch and saw the identical conflicts.

PR #406, opened 2023-12-23
add boxed union feature to circumvent stack overflow

rexmas:
For enums with a lot of variants the `Strategy` synthesized by the `#[derive(Arbitrary)]` macro is a right-deep `TupleUnion`.

This commit adds a `boxed_union` feature, which when enabled will change the `enum` expansion to use `Union<BoxedStrategy<Self>>`.

see github issues:
https://github.com/proptest-rs/proptest/issues/249 https://github.com/proptest-rs/proptest/issues/152

code included from https://github.com/proptest-rs/proptest/issues/249#issuecomment-1367820389

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -70,3 +70,6 @@ jobs:
       - name: Clean and Run tests for derive
         if: ${{ matrix.build == 'nightly' }}
         run: cd proptest-derive && cargo clean && cargo test
+      - name: Clean and Run tests for derive with features enabled
+        if: ${{ matrix.build == 'nightly' }}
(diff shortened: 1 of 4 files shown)


matthew-russo (approved):
Haven't pulled it locally and tested it but lgtm.

matthew-russo:
fixes #152 (sort of. not as described in that issue)

i wouldn't mark this on as fixing #249 as I see that one being more about the internals of ValueTree itself. unless i'm misunderstanding something

rbtcollins:
It fixes 249 as I intended to write it! With this, my openapi struct would be generated to heap, rather than stack, which would permit running the tests without --release.

PR #418, opened 2024-01-13
Fix re-complication of initial state

timstobal:
This PR fixes an issue where, after the initial state in a state machine test has been simplified, it will not be re-complicated when this results in a successful run. As a result, when this situation is encountered, the test will report a "minimal failing input" that is actually a successful input - the input that is just barely simpler than the minimal failing input.

I've also added a proptest that detects the failure, along with two regression seeds, for good measure.


matthew-russo (approved):
thanks for the fix!

tzemanovic (approved):

PR #419, opened 2024-01-14
Add features `handle-panics` and `backtrace`

target-san:
* `handle-panics` attempts to capture panic backtrace when one happens, instead of spitting it to console
* `backtrace` enables actual backtraces using `std::backtrace`
* Short backtrace printed second time after test failure is produced by toplevel test func reporting its failure via panic, can be remedied by rewriting helper macros so that they return failure instead of panicking.
* Collected backtrace is still 50+ frames long. Unfortunately it seems to be impossible to even enumerate frames on stable, yet alone use existing short backtrace style. `Backtrace::frames()` and `BacktraceFrame` are still experimental.

### DIFF
--- a/proptest-derive/CHANGELOG.md
+++ b/proptest-derive/CHANGELOG.md
@@ -1,8 +1,14 @@
 ## Unreleased
 
+### Features
+
 - Add `boxed_union` feature which when turned on uses heap allocation for
   `#[derive(Arbitrary)]` strategy synthesis preventing stack overflow for
   exceptionally large structures.
+
+### Dependencies
+
+- Upgraded `syn` to 2.x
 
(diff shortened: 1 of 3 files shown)


target-san:
Note.

I'm aware that collected backtrace is still 50+ frames long. Unfortunately it seems to be impossible to even enumerate frames on stable, `Backtrace::frames()` and `BacktraceFrame` are still experimental.

Default test harness also prints short backtrace from the point `proptest`'s test runner panicked with final result. It's possible to avoid this behavior by making generated test methods return return `Result<(), Whatever>` (edition 2018), yet requires significant rewrites in helper macros. Is this acceptable or desirable?

PR #420, opened 2024-01-20
Fix hygiene issues with prop_state_machine macro

timstobal:
This PR eliminates unqualified type names from the prop_state_machine macro, allowing it to be used without importing all of the required types.

This is a minor quality of life improvement, as it eliminates a class of compile errors, the cause of which is not immediately obvious from the perspective of the author of a test.

A pair of no-op unit tests are also included in a module with no imports, to verify at test compile time that the macros are hygienic.


matthew-russo (approved):
lgtm. thanks for the fix

tzemanovic (approved):

PR #421, opened 2024-01-21
Add features `handle-panics` and `backtrace` 

### LINKED ISSUE #356: add backtraces in final error report
On a test failure after shrinking, proptest error report doesn't include a backtrace. It can be found further back from the last run that's panicked, but when a test also prints during its execution, this can get scrolled out of view. It is possible to capture the backtrace with a panic hook to then report it together with the minimal failed input.

target-san:
* `handle-panics` attempts to capture panic backtrace when one happens, instead of spitting it to console
* `backtrace` enables actual backtraces using `std::backtrace`
* Short backtrace printed second time after test failure is produced by toplevel test func reporting its failure via panic, can be remedied by rewriting helper macros so that they return failure instead of panicking.
* Collected backtrace is still 50+ frames long. Unfortunately it seems to be impossible to even enumerate frames on stable, yet alone use existing short backtrace style. `Backtrace::frames()` and `BacktraceFrame` are still experimental.

P.S: Sorry for "PR spam", prev one #419 was closed prematurely due to accident

closes #356

### DIFF
--- a/proptest/src/test_runner/errors.rs
+++ b/proptest/src/test_runner/errors.rs
@@ -138,10 +138,17 @@ impl<T: fmt::Debug> fmt::Display for TestError<T> {
             TestError::Abort(ref why) => write!(f, "Test aborted: {}", why),
             TestError::Fail(ref why, ref bt, ref what) => {
                 writeln!(f, "Test failed: {}.", why)?;
+
                 if !bt.is_empty() {
-                    writeln!(f, "{bt}")?;
+                    writeln!(f, "\nstack backtrace:\n{bt}")?;
+                    // No need for extra newline, backtrace seems to print it anyway
+                } else {
+                    // Extra empty line between failure description and minimal failing input
+                    writeln!(f)?;
                 }
-                write!(f, "minimal failing input: {:#?}", what)
+
+                writeln!(f, "minimal failing input: {:#?}", what)?;
+                Ok(())
             }
         }
     }
--- a/proptest/src/test_runner/errors.rs
+++ b/proptest/src/test_runner/errors.rs
@@ -125,7 +125,8 @@ impl<T: PartialEq> PartialEq for TestError<T> {
             (Self::Fail(l0, _, l2), Self::Fail(r0, _, r2)) => {
                 l0 == r0 && l2 == r2
             }
-            _ => false,
+            (TestError::Abort(_), TestError::Fail(_, _, _))
+            | (TestError::Fail(_, _, _), TestError::Abort(_)) => false,
         }
     }
 }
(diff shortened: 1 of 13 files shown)


proptest/src/test_runner/runner.rs
```diff
@@ -210,6 +212,96 @@ where
...
+
+    thread_local! {
+        // Pointers to arbitrary fn's are fat, and Rust doesn't allow crafting null pointers
+        // to fat objects. So we just store const pointer to tuple with whatever data we need
+        static HANDLER: Cell<*const (*mut dyn FnMut(&PanicInfo<'_>),)> = Cell::new(ptr::null());
+    }
+
+    static INIT_ONCE: Once = Once::new();
+    // NB: no need for external sync, value is mutated only once, when init is performed
+    static mut DEF_HOOK: Option<Box<dyn Fn(&PanicInfo<'_>) + Send + Sync>> =
```
rexmas, line 231: please expand this to `DEFAULT_HOOK` for clarity, i'm concerned that "DEF" could be mistaken for something else

proptest/src/test_runner/runner.rs
```diff
@@ -210,6 +212,96 @@ where
     })
 }
 
+#[cfg(feature = "handle-panics")]
+mod panicky {
```
rexmas, line 216: given the unsafe usage, this section could use a bit more documentation i think. possibly specifying what the responsibility of each static variable and each function as well as clarifying the order of operations. i don't think it needs to be long but just enough to provide more clarity

proptest/src/test_runner/errors.rs
```diff
@@ -99,24 +107,48 @@ impl<E: ::std::error::Error> From<E> for TestCaseError {
...
 }
 
+impl<T: PartialEq> PartialEq for TestError<T> {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Abort(l0), Self::Abort(r0)) => l0 == r0,
+            (Self::Fail(l0, _, l2), Self::Fail(r0, _, r2)) => {
+                l0 == r0 && l2 == r2
+            }
+            _ => false,
```
rexmas, line 128: i think i'd prefer if we exhaustively match here, even if it's more verbose

rexmas (commented):
thank you for this pr, i think this will be a very nice addition :)

proptest/src/test_runner/runner.rs
target-san: Done

rexmas (approved):
thank you for the detailed docs, this looks great 👍

rexmas:
cc @matthew-russo

matthew-russo:
I had started going through this last night. will finish up later tonight

target-san:
Ok, then I guess the best course is to take out `"backtrace"` feature and try introduce it as separate PR. Will do it in a few days.

matthew-russo:
we're looking at a 2.0 release so we can include this and a couple other things that are pending. it'd probably take 1-2 months to get everything together if those timelines work for you. also happy to accept changes with no backwards compatibility issues in the meantime if you want a faster turnaround

matthew-russo:

PR #424, opened 2024-02-18
[Fix] Config : don't clobber existing failure-persistence config during contextualization

### LINKED ISSUE #423: Unable to set `failure_persistence` using `#![proptest_config]`
I have a project where I want to change the `failure_persistence` configuration, but it does not seem to be taking effect. When I run the following test, it still shows the `failed to find lib.rs or main.rs` warning 

```
running 1 test
proptest: FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs
29
24
14
test test::test ... ok
```

despite the `failure_persistence` being set to `FileFailurePersistence::Direct`. The `cases` property in the same configuration is being picked up.

`src/test.rs`:
```rust
fn main() {
    println!("Hello, world!");
}

#[cfg(test)]
mod test {
    use proptest::test_runner::FileFailurePersistence;

    proptest::proptest! {
        #![proptest_config(proptest::test_runner::Config {
            failure_persistence: Some(Box::new(FileFailurePersistence::Direct("failure_persistence.txt"))),
            cases: 3,
            ..Default::default()
        })]
        #[test]
        fn test(a in 0..30) {
            println!("{a}")
        }
    }
}
```

`Cargo.toml`:
```
[package]
name = "proptest-file"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[[bin]]
name = "test"
path = "src/test.rs"

[dependencies]
proptest = "1.4.0"
```

matthew-russo:
fixes #423

### DIFF
--- a/proptest/src/test_runner/config.rs
+++ b/proptest/src/test_runner/config.rs
@@ -80,8 +80,6 @@ pub fn contextualize_config(mut result: Config) -> Config {
         }
     }
 
-    result.failure_persistence =
-        Some(Box::new(FileFailurePersistence::default()));
     for (var, value) in
         env::vars_os().filter_map(|(k, v)| k.into_string().ok().map(|k| (k, v)))
     {
@@ -186,8 +184,11 @@ fn default_default_config() -> Config {
 // defaults.
 #[cfg(feature = "std")]
 lazy_static! {
-    static ref DEFAULT_CONFIG: Config =
-        contextualize_config(default_default_config());
+    static ref DEFAULT_CONFIG: Config = {
+        let mut default_config = default_default_config();
+        default_config.failure_persistence = Some(Box::new(FileFailurePersistence::default()));
+        contextualize_config(default_config)
+    };
 }
 
 /// Configuration for how a proptest test should be run.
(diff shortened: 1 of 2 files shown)


proptest/src/test_runner/config.rs
```diff
@@ -80,8 +78,6 @@ pub fn contextualize_config(mut result: Config) -> Config {
         }
     }
 
-    result.failure_persistence =
-        Some(Box::new(FileFailurePersistence::default()));
```
rexmas, line 84: seems like this was in some sense the default for the `std` feature? does it make sense adjust `DEFAULT_CONFIG` to be setup with `Some(Box::new(FileFailurePersistence::default()))` if feature is `std`?

matthew-russo: fair point. will update the `Default` impl

rexmas (approved):
nice 👍

ergl:
@matthew-russo Would it be possible to cut a release with this fix? Still running into this problem with 1.4.0

matthew-russo:
yes i need to cut a release for derive as well. will do that later today

matthew-russo:
sorry for the delay -- these are published on crates.io now

PR #432, opened 2024-03-06
Use `array::from_fn` instead of `unarray::build_array`

DaniPopes:
`core::array::from_fn` is stable since 1.63, and the current MSRV is 1.64.

Ideally `unarray` would get entirely replaced but `core::array::try_from_fn` is
still unstable.


matthew-russo (approved):
thanks

PR #457, opened 2024-06-04
config: properly read environment

bmwill:
I noticed that when deriving `Arbitrary` for a type that contains a `Vec`, and then using it in a proptest generally led to panics with the range or size of the generated Vec being empty or `0..0`. After some digging around (and actually reading some of the output warnings when running a test) I realized that the config initialization logic wasn't working as intended. This patch should address it.


matthew-russo:

PR #458, opened 2024-06-04
fix wasm support

bmwill:
This is a rebased version of #270 that I needed to get proptest working in wasm environments.


matthew-russo:

Issue #453, opened 2024-05-23
Change verbosity of shrink disabled and persisted failure messages 


grandizzy:
I am working to implement an (indicatif based) progress bar for tests that use proptest and I need to suppress couple of messages that I really think should follow the info verbose rule (as they have to be explicitly configured, so not unexpected errors per se). Please let me know if you'd be open for a PR with changes below. Thank you!
 
- when shrink is disabled: do not start loop and exit early in `shrink` fn
https://github.com/proptest-rs/proptest/blob/c117b769ba90e0e70c7fa020c29da9aaf89af493/proptest/src/test_runner/runner.rs#L754-L763
with a info message (since disabling should be explicitly set in Config as `max_shrink_iters: 0`), proposed addition :

```diff
         fork_output: &mut ForkOutput,
         is_from_persisted_seed: bool,
     ) -> Option<Reason> {
+        if self.config.max_shrink_iters() == 0 {
+            verbose_message!(
+                self,
+                INFO_LOG,
+                "Shrinking disabled by configuration"
+            );
+            return None
+        }
+
```

- when failure is persisted: apply same `INFO_LOG` level to persist message below (since `failure_persistence` should explicitly be set in Config for recording failures)
https://github.com/proptest-rs/proptest/blob/c117b769ba90e0e70c7fa020c29da9aaf89af493/proptest/src/test_runner/failure_persistence/file.rs#L154-L161

grandizzy:
ping @matthew-russo any insights highly appreciated. thank you!

matthew-russo:
Sorry for the delay, I had to step away for the past couple months to deal with some things at home. I'll be back working on things starting this weekend and will start triaging the issues that have come in.

I think the first suggestion is non-controversial and easy to accept.

I think the second suggestion isn't as straightforward since its only logging because there was a failure (hence its in the context of an error) and its anticipating that it may be running in a CI system, where users may only keep stdout. How are you going about suppressing it right now?

grandizzy:
Thank you! There's no suppression rn so basically the progress bar gets flaky when proptest prints messages. I will make the PR for first case, what if for 2nd one we add an env setting that if turned on (default off) would skip printing such messages?

PR #459, opened 2024-06-10
Verbosity of shrink disabled and persisted failure messages

grandizzy:
re #453 handle exit early with INFO message
To be determined best way to handle the failure persisted message
https://github.com/proptest-rs/proptest/issues/453#issuecomment-2155796541


matthew-russo:

PR #461, opened 2024-06-10
feat: add possibility to interrupt running test

### LINKED ISSUE #460: feat: add option to interrupt test runner
At this moment there's no way to programmatically stop a test run in process with replay, which simply loops until a failure or all configured cases pass
https://github.com/proptest-rs/proptest/blob/24412f5f10a365c8f12a71e3b3a053677ced91b0/proptest/src/test_runner/runner.rs#L609

It would be nice to be able to interrupt this loop at will so in a setup with multiple tests started at the same time (not using fork feature but rather parallelized using rayon for example) following scenarios could be achieved:
- fail fast: when one of these test fails then the other tests are interrupted
- timeout (not using rusty fork / fork timeout feature): interrupt all running tests when timer (started outside tests) expires

grandizzy:
Closes #460 
- add `Interrupt` test case error: it doesn't count as test failure but signals to stop executing new runs in given test runner
- if test result is interrupt then break the loop to configured successful test cases that must execute
- doesn't count as a failure
- tests

### DIFF
--- a/proptest/src/test_runner/errors.rs
+++ b/proptest/src/test_runner/errors.rs
@@ -25,6 +25,8 @@ use crate::test_runner::Reason;
 /// `Error::display()` into the `Fail` case.
 #[derive(Debug, Clone)]
 pub enum TestCaseError {
+    /// The test runner was interrupted.
+    Interrupt(Reason),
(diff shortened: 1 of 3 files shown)


matthew-russo:

Issue #449, opened 2024-05-18
Panic on empty ranges during strategy creation


mirandaconrado:
I have encountered scenarios in my code where I end up with empty ranges. Most of the time this happens when one of the bounds of the range comes from an arbitrary value. It's kind of hard to track these down because they fail at the moment that values are generated instead of failing when the strategy is created, even though we know by then that it will fail.

Example:
```
use proptest::prelude::*;

proptest! {
    #[test]
    fn empty_range(s in 0..0) {}
}
```

causes

```
---- empty_range stdout ----
thread 'empty_range' panicked at /Users/conrado/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/distributions/uniform.rs:561:1:
Uniform::new called with `low >= high`
stack backtrace:
   0: rust_begin_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:144:5
   3: <rand::distributions::uniform::UniformInt<i32> as rand::distributions::uniform::UniformSampler>::new
             at /Users/conrado/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/distributions/uniform.rs:452:17
   4: rand::distributions::uniform::Uniform<X>::new
             at /Users/conrado/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/distributions/uniform.rs:189:17
   5: proptest::num::sample_uniform
             at /Users/conrado/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proptest-1.4.0/src/num.rs:28:5
   6: proptest::num::i32::<impl proptest::strategy::traits::Strategy for core::ops::range::Range<i32>>::new_tree
             at /Users/conrado/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proptest-1.4.0/src/num.rs:74:21
   7: <proptest::strategy::map::Map<S,F> as proptest::strategy::traits::Strategy>::new_tree
[...]

PR #462, opened 2024-06-13
Detect empty ranges during tree creation

mirandaconrado:
Currently, if an empty range is used as base for a strategy, the generation will fail during the call to the `rand` crate as it cannot sample from an empty range. This PR creates a panic during the tree construction (the soonest we can) with a descriptive message, facilitating tracking the source of the issue.

Solves https://github.com/proptest-rs/proptest/issues/449

The panic message test pattern follows Rust's core library https://github.com/rust-lang/rust/blob/a1bad57fa59a8069a6ebb05cd6a2ae73c88b2e98/src/test/run-pass/panic-uninitialized-zeroed.rs


matthew-russo (approved):

rexmas (approved):
great! thanks for the contribution!

PR #464, opened 2024-06-15
[Doc] Arbitrary : add a note about the derive macro in Arbitrary's documentation

### LINKED ISSUE #463: derive macro for Arbitrary is hard to find
probably the biggest hurdle to using proptest is the need to manually implement Arbitrary for a type.  the ability to have a derive macro, even if it was really limited (eg. only unit enums) would be incredibly useful.

obviously this derive macro couldn't figure out the best methods for shrinking a value and such, but often it wouldn't need to, as a lot of types (like unit enums) don't have any meaningful way they can be shrunk.

matthew-russo:
fixes #463


rexmas (approved):
👍

PR #469, opened 2024-06-22
[proptest-macro] [Fix] Manifest : add metadata required to publish, copied from core proptest crate

matthew-russo:


rexmas (approved):
👍

cameron1024 (approved):

Issue #161, opened 2019-07-22
Idea: Provide print/dbg macros that only print values of minimal examples


Andlon:
I'm thoroughly enjoying using `proptest`, but there's in particular one issue that makes it perhaps a bit harder than necessary to actually debug failing tests. When using e.g. `println!` or `dbg!` in a proptest, the output is printed for every test that `proptest` runs. It would be very, very practical if one could print only for the (minimal/shrunk) failing example. Then you could even leave these print/debug statements in the test as they anyway wouldn't be printing anything as long as the test is passing.

The C++ library `rapidcheck` has a similar feature. I imagine something like:

```rust
proptest! {
    #[test]
    fn test_whatever(value in strategy()) {
        prop_dbg!(&value);
        prop_println!("Value is: {:?}", value);
        prop_assert!(has_property(value));
    }
}
```

Perhaps there are also other workflows for debugging failing tests? Sometimes it's just excruciatingly painful to reconstruct the failing example in code and make a separate test for it.

Andlon:
As a follow up, it would perhaps be of far more general use if there was a way to invoke code only for the minimal counter-example. Something like

```rust
#[test]
    fn test_whatever(value in strategy()) {
        if minimal_witness {
            print_lots_of_debug_stuff(&value);
        }
        prop_assert!(has_property(value));
    }
```

Andlon:
I keep running into exactly this issue. For example, right now I have a failing test, and I'd like to print values only for that minimal test. I'm forced to put `dbg!` or similar in my main test body, which means my console is cluttered with output for thousands of tested cases even though I'm only interested in the minimal witness. It's also not clear to me if the "last run" example truly is the minimal example. 

Is this really not a problem that others regularly run into?

Here are two more concrete suggestions for how to improve this situation:

- Augment the `test` closure in [`TestRunner::run`](https://docs.rs/proptest/1.2.0/proptest/test_runner/struct.TestRunner.html#method.run) to additionally take a boolean parameter indicating whether the value provided to the closure is a minimal witness. This is a _breaking change_.
- Alternatively, a way to provide this functionality without changing the API could be to use a thread-local boolean is set before the test is run, and which can be accessed through e.g. a macro `prop_is_minimal!().

In both cases, the syntax inside of a test *might* look like something like

```rust
if prop_is_minimal!() {
  // Print stuff, export debug/visualization info to files etc.
}
```

I'm not otherwise familiar with the code base, so perhaps there are much better solutions available.

[...]

PR #477, opened 2024-07-05
Implement is_minimal_case

Andlon:
This is a draft for an implementation of the idea suggested in #161.

Currently, using `proptest` for debugging is very cumbersome because you cannot easily run code only for the minimal failing test case. In my experience, I usually only care about the minimal test case, but if I want to print some debug info about that failing test case, I'll have to print for *all* test cases, which is borderline useless.

This PR introduces a method `is_minimal_case` that you can call inside a `proptest` to determine if the current test case is the minimal failing case. To do this, the minimal test case is run one more time after it has been identified, and a thread local boolean is set before and after.

This allows you to write tests like these:
```rust
use proptest::{proptest, prop_assert, is_minimal_case};

proptest! {
    #[test]
    fn test_is_not_five(num in 0 .. 10) {
        if is_minimal_case() {
            eprintln!("Minimal test case is {num:?}");
            export_to_file_for_analysis(num);
        }

        prop_assert!(num != 5);
    }
}
```

A nice side effect of this is that it makes it much easier to use `proptest` with a debugger: Simple add an `if` condition like the one above, and put a breakpoint on a statement inside the branch. Then you can run the debugger only through the minimal failing test case.

I haven't yet written any tests, because I'm frankly not quite sure how best to go about it. I suppose at a minimum we should have a test where a `TestRunner` is manually invoked with a dummy test case, where we test that the reported minimal test case is what we expect. But I'm not sure how best to write such a test in a way that fits well with the rest of the code base. I'd be very grateful to have some suggestions here.

I'm not sure how to correctly handle forking here - if anything in particular even needs to be done differently. And whether or not I should forward `is_from_persistent_seed`. Since I'm not so familiar with the code base I don't really understand the ramifications.

Happy to make any necessary adjustments to the PR.

### DIFF
--- a/proptest/src/is_minimal_case.rs
+++ b/proptest/src/is_minimal_case.rs
@@ -37,7 +37,7 @@ thread_local! {
 /// }
 /// ```
 pub fn is_minimal_case() -> bool {
-    IS_MINIMAL_CASE.get()
+    IS_MINIMAL_CASE.with(|cell| cell.get())
 }
 
 /// Helper struct that helps to ensure panic safety when entering a minimal case.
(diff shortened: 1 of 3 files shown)


Andlon:
I have no idea what happened with CI. It seems that it just hangs on some build script?

matthew-russo:

PR #482, opened 2024-07-10
fix: check preconditions on correct state

thomaseizinger:
When shrinking a state machine test, proptest either shrinks a transition, deletes a transition or attempts to shrink the initial state. The `check_acceptable` function however uses the `last_valid_initial_state` which isn't the correct one if we just shrank / complicated the initial state.

### DIFF
--- a/proptest-state-machine/src/strategy.rs
+++ b/proptest-state-machine/src/strategy.rs
@@ -358,7 +358,9 @@ impl<
                 DeleteTransition(ix - 1)
             };
             // If this delete is not acceptable, undo it and try again
-            if !self.check_acceptable(None) {
+            if !self
+                .check_acceptable(None, self.last_valid_initial_state.clone())
+            {
                 self.included_transitions.set(ix);
                 self.last_shrink = None;
                 return self.try_simplify();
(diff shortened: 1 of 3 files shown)


proptest-state-machine/src/strategy.rs
```diff
@@ -606,7 +617,7 @@ impl<
             }
             Some(InitialState) => {
                 if self.initial_state.complicate()
-                    && self.check_acceptable(None)
```
thomaseizinger, line 609: This is the bug (in my opinion). Previously, we checked whether the **last** valid state was acceptable but when we are shrinking the `initial_state`, we need to check whether the current one is acceptable.

proptest/src/test_runner/runner.rs
thomaseizinger, line 1: Happy to split these out into a separate PR if you'd like. I found them helpful during debugging.

thomaseizinger:
@matthew-russo Hopefully you don't mind the ping! Would appreciate if you could have a look at this and let me know your opinion on it! :)

matthew-russo (approved):
Seems okay to me but I've added Tomáš as well to double check

tzemanovic:
Nice catch, thank you!

tzemanovic (approved):

nickelization:
I independently ran into this same bug the other day, and once I figured out it was a proptest bug I was happy to see it's already been fixed 😄 Curious if there are plans to cut a new release anytime soon, I'd love to have this fix in a published version of the crate.

tzemanovic:
> I independently ran into this same bug the other day, and once I figured out it was a proptest bug I was happy to see it's already been fixed 😄 Curious if there are plans to cut a new release anytime soon, I'd love to have this fix in a published version of the crate.

apologies for the long delay, will prepare a patch release today

tzemanovic:
released in v0.3.1

PR #491, opened 2024-08-11
docs: Stragegy -> Strategy in rustdoc for Union

benjaminedwardwebb:
Fix a typo in the rustdoc for the proptest::strategy::Union struct.


rexmas (approved):
thanks!

PR #493, opened 2024-08-16
feat: introduce `Sequential::new`

thomaseizinger:
The `Sequential` struct is where most of the useful functionality of `proptest-state-machine` lives. Currently, it can only be constructed via the `ReferenceStateMachine` trait.

By introducing a dedicated ctor function, users are not bound to use the `ReferenceStateMachine` trait but can initialise the test to their liking whilst still being able to use all of the shrinking functionality built into `Sequential`.


thomaseizinger:
@matthew-russo Thoughts on this one?

matthew-russo:
Sorry i had missed this. lgtm -- approved new version and merged

PR #498, opened 2024-08-24
remove unused 'frunk' feature and add From's for SizeRange and Probability

### LINKED ISSUE #492: How to convert `SizeRange` to a regular `Range`?
Using start/end/end_incl/... manually can lead to errors.
Found `frunk::Generic` trait implementation for `SizeRange`, but feature is not documented.
I use my own function. Sad.

matthew-russo:
fixes #492


rexmas (approved):
makes sense, nice cleanup!

PR #499, opened 2024-08-26
Make libm optional in a std environment

bantonsson:
When adding `proptest` as a `dev-dependency` it will add `libm` transitively no matter what, and this has noticeable effect on our criterion benchmarks.

This happens even if you add `proptest` like this:

```
proptest = {version = "1.5", features = ["std"], default-features = false}
```


matthew-russo (approved):

matthew-russo:

PR #511, opened 2024-09-22
Fix bounds generation for generics in derive(Arbitrary)

matthew-russo:
The implementation of UseTracker expects that iteration over items of used_map gives items in insertion order. However, the order of BTreeSet is based on Ord, not insertion.

I use a Vec of tuples. An alternative would be indexmap crate, but since the maps are supposed to be very small, it is probably not worthy. There already was a mention about the crate in the comments worrying about compile times.

replaces #260


rexmas (approved):
nice find 👍

PR #517, opened 2024-09-22
[proptest] Minimal implementation of interceptor for excessive panics outputs

matthew-russo:
replaces #421


target-san:
#525 is a part of this PR which can be added without breaking changes. I'll add backtraces part as separate PR

PR #519, opened 2024-09-22
458 fix wasm

matthew-russo:
replaces #458

### DIFF
--- a/proptest/src/test_runner/config.rs
+++ b/proptest/src/test_runner/config.rs
@@ -10,50 +10,13 @@
 use crate::std_facade::Box;
 use core::u32;
 
-#[cfg(feature = "std")]
-use std::env;
-#[cfg(feature = "std")]
-use std::ffi::OsString;
-#[cfg(feature = "std")]
-use std::fmt;
-#[cfg(feature = "std")]
-use std::str::FromStr;
-
 use crate::test_runner::result_cache::{noop_result_cache, ResultCache};
 use crate::test_runner::rng::RngAlgorithm;
 use crate::test_runner::FailurePersistence;
-#[cfg(feature = "std")]
-use crate::test_runner::FileFailurePersistence;
-
-#[cfg(feature = "std")]
-const CASES: &str = "PROPTEST_CASES";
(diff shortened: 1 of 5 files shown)


RReverser:
From a quick glance, looks good, although it's been open for a long time, so would be good to test it still works.

proptest/src/test_runner/config.rs
```diff
@@ -10,51 +10,34 @@
...
-#[cfg(feature = "std")]
-const MAX_GLOBAL_REJECTS: &str = "PROPTEST_MAX_GLOBAL_REJECTS";
-#[cfg(feature = "std")]
-const MAX_FLAT_MAP_REGENS: &str = "PROPTEST_MAX_FLAT_MAP_REGENS";
-#[cfg(feature = "std")]
-const MAX_SHRINK_TIME: &str = "PROPTEST_MAX_SHRINK_TIME";
-#[cfg(feature = "std")]
-const MAX_SHRINK_ITERS: &str = "PROPTEST_MAX_SHRINK_ITERS";
-#[cfg(feature = "std")]
-const MAX_DEFAULT_SIZE_RANGE: &str = "PROPTEST_MAX_DEFAULT_SIZE_RANGE";
```
rexmas, line 41: did this one get deleted on accident? i believe it's still in use in `parse_or_warn`

RReverser: Idk - it's been a very long time - but CI seems to pass?

bmwill: Yeah this is fixed in https://github.com/proptest-rs/proptest/pull/518

RReverser: > so arbitrary tokens in pattern match positions seem to be treated as new variable bindings and default to (), that's why this still compiles even though there's a warning.

Hm but that doesn't explain why `parse_or_warn` accepted it if it's `()` and used in an argument that expects `&str`.

matthew-russo: pushed an updated commit that has this

rexmas (approved):
nice

PR #523, opened 2024-09-22
[proptest-macro] add the ability to specify custom strategies

matthew-russo:
replaces #478

### DIFF
--- a/proptest-macro/src/property_test/tests/snapshot_tests.rs
+++ b/proptest-macro/src/property_test/tests/snapshot_tests.rs
@@ -1,14 +1,47 @@
-use syn::{parse_str, ItemFn};
+use syn::parse_quote;
 
 use crate::property_test::{codegen, options::Options};
 
-#[test]
-fn basic_derive_example() {
-    let f: ItemFn =
-        parse_str("fn foo(x: i32, y: String) { let x = 1; }").unwrap();
-    let tokens = codegen::generate(f, Options::default());
-    let file = syn::parse_file(&tokens.to_string()).unwrap();
-    let formatted = prettyplease::unparse(&file);
+/// Helper macro to make snapshot tests
+///
+/// The code inside the block is parsed as a function and the `#[property_test]` macro is applied
+/// to it. The generated code is then formatted, and passed to the snapshot testing library
+///
+/// If `fails` is supplied, then the output will be
+macro_rules! snapshot_test {
+    ($name:ident {$($t:tt)*}) => {
+        #[test]
+        fn $name() {
+            let input = parse_quote! { $($t)* };
+            let tokens = codegen::generate(input, Options::default());
+            let file = syn::parse_file(&tokens.to_string()).unwrap();
+            let formatted = prettyplease::unparse(&file);
 
-    insta::assert_snapshot!(formatted);
+            insta::assert_snapshot!(formatted);
+        }
+    };
 }
+
+snapshot_test!(basic_derive_example {
+    fn foo(x: i32, y: String) {
+        let x = 1;
+    }
+});
+
+snapshot_test!(custom_strategy {
+    fn foo(
+        #[strategy = 123] x: i32,
+        #[strategy = a + more()("complex") - expression!()] y: String,
+    ) {
+        let x = 1;
+    }
+});
+
+snapshot_test!(mix_custom_and_default_strategies {
+    fn foo(
+        x: i32,
+        #[strategy = a + more()("complex") - expression!()] y: String,
+    ) {
+        let x = 1;
+    }
+});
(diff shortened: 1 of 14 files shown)


proptest-macro/src/property_test/validate.rs
```diff
@@ -26,34 +29,96 @@ fn all_args_non_self(f: &ItemFn) -> Result<(), TokenStream> {
...
+        let old_attrs = std::mem::take(&mut pat_ty.attrs);
+
+        // every strategy attr should have the form `#[strategy = <expr>]`
+        for attr in old_attrs.into_iter().filter(is_strategy) {
+            match attr.meta {
+                // a "good" strategy - if we see more than one, emit an error
+                Meta::NameValue(_) => {
+                    if first_strategy_seen {
+                        error.extend(quote_spanned! {
+                            attr.span() => compile_error!("duplicate `#[strategy = ...]` definition");
```
rexmas, line 59: may be worth saying the name of the parameter here for added clarity

proptest-macro/src/property_test/tests/snapshot_tests.rs
```diff
@@ -1,14 +1,47 @@
...
+snapshot_test!(basic_derive_example {
+    fn foo(x: i32, y: String) {
+        let x = 1;
+    }
+});
+
+snapshot_test!(custom_strategy {
+    fn foo(
+        #[strategy = 123] x: i32,
+        #[strategy = a + more()("complex") - expression!()] y: String,
```
rexmas, line 34: this is really cool that one can just assign a literal if needed. i'm curious about `y` here though, like what happens if an operator that cannot be used on string types is used? I guess that question may reduce down to, what happens if the strategies type does not match the type of the parameter, will the type error be relatively obvious for the user?

rexmas: also do we expect argument types to every use lifetimes or generics? what happens if someone tries to and should we test for that?

rexmas (approved):
had a few questions but overall this looks great to me and happy to move forward with it! (assuming ci is 💚 after running)

also just to weigh in a comment from the original pr
>I think for now, disallowing other attributes is safer from a backwards compatibility point of view - we don't lock ourselves into supporting something that is difficult later

i think i agree with taking a conservative/iterative stance here

rexmas:
cc @cameron1024

proptest-macro/src/property_test/validate.rs
cameron1024: Worth mentioning that, since we're using `quote_spanned`, the compile error will appear in IDEs to be "coming from" the second attribute (i.e. that's where the red squiggly line will appear)

proptest-macro/src/property_test/tests/snapshot_tests.rs
cameron1024: Sorry I've had these comments stuck in "pending" :facepalm: 

 - If there's an error typechecking the expression itself (e.g. `1 + "hello"`, we'll get a nice rustc error:
```
error[E0277]: cannot add `&str` to `{integer}`
  --> src/main.rs:29:27
   |
29 |     fn foo(#[strategy = 1 + "hello"] x: bool) {
   |                           ^ no implementation for `{integer} + &str`
   |
```
 - if the expression has a type where the `Strategy` impl has the wrong `Value` type (or it doesn't implement `Strategy` at all) we also get a nice-ish error:
```
rror[E0599]: the method `prop_map` exists for tuple `(NotAStrategy,)`, but its trait bounds were not satisfied
  --> src/main.rs:30:5
   |
22 | struct NotAStrategy;
   | ------------------- doesn't satisfy `NotAStrategy: proptest::strategy::Strategy`
...
30 |     #[proptest::property_test]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called on `(NotAStrategy,)` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `NotAStrategy: proptest::strategy::Strategy`
           which is required by `(NotAStrategy,): proptest::strategy::Strategy`
           `(NotAStrategy,): proptest::strategy::Strategy`
           which is required by `&(NotAStrategy,): proptest::strategy::Strategy`
           `(NotAStrategy,): proptest::strategy::Strategy`
           which is required by `&mut (NotAStrategy,): proptest::strategy::Strategy`
```
 - we can have generics, but they have to be used with a concrete instantiation (i.e. you can have `Foo<i32>` but not `Foo<T>`. not sure how you'd do lifetimes, but I'm not 100% sure what that would even mean, since test functions can't have lifetimes. But the following works:
```rust
struct Foo<T>(PhantomData<T>);

impl Strategy for Foo<i32> {
  type Value = i32;
  // ...
}

#[property_test]
fn foo(#[strategy = Foo<i32>] x: i32) {}
```

PR #525, opened 2024-10-05
Add feature `handle-panics`

target-san:
When `handle-panics` is enabled, panic backtrace which happens in test case is silenced instead of spitting it to console
This PR is part of #421/#517 which doesn't change public API. Part with backtraces capture will be presented as separate PR

P.S: Do we actually need feature to gate this out? Or should it be just enabled by `std`?

### DIFF
--- a/proptest/src/test_runner/scoped_panic_hook.rs
+++ b/proptest/src/test_runner/scoped_panic_hook.rs
@@ -1,3 +1,12 @@
+//-
+// Copyright 2024 The proptest developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
 #[cfg(feature = "handle-panics")]
 mod internal {
     //! Implementation of scoped panic hooks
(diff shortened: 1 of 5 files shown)


proptest/src/test_runner/scoped_panic_hook.rs
```diff
@@ -0,0 +1,132 @@
...
+        let handler = SCOPED_HOOK_PTR.get();
+        if !handler.is_null() {
+            // It's assumed that if container's ptr is not null, ptr to `FnMut` is non-null too.
+            // Correctness **must** be ensured by hook switch code in `with_hook`
+            let hook = unsafe { &mut *(*handler).0 };
+            (hook)(info);
+            return;
+        }
+
+        if let Some(hook) = unsafe { DEFAULT_HOOK.as_ref() } {
```
rexmas, line 67: I think we might want to add `#[allow(static_mut_refs)]` right above this to eliminate that warning.

rexmas (approved):
looks good, i'm in favor of having a feature flag for this for now since some people may rely on the output of panics still and not want them suppressed. when we ship the other pr in i may be in favor of removing the ff since we'll need a major version bump

matthew-russo (approved):

proptest/src/test_runner/scoped_panic_hook.rs
target-san: I think it's worth more to run full fmt+clippy check as a separate PR. There are too many warnings besides this one.

PR #526, opened 2024-10-06
Add feature `backtrace`

target-san:
! Merge after #525

Enables collection of full backtraces:
* when panic occurs inside test, but only if `handle-panics` feature is enabled
* when `prop_assert` macro or any of its derivatives is used

### DIFF
--- a/proptest/src/test_runner/reason.rs
+++ b/proptest/src/test_runner/reason.rs
@@ -151,7 +151,7 @@ impl<'a> fmt::Display for DisplayReason<'a> {
         if bt.is_empty() {
             write!(f, "{msg}")
         } else {
-            write!(f, "{msg}\n{bt}")
+            write!(f, "{msg}\nstack backtrace:\n{bt}")
         }
     }
 }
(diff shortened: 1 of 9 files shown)


proptest/src/test_runner/reason.rs
```diff
@@ -7,48 +7,201 @@
...
 ///
-/// Currently this is merely a wrapper around a message, but more properties
-/// may be added in the future.
+/// Contains message which describes reason and optionally backtrace
+/// (depending on several factors like features `backtrace` and
+/// `handle-panics`, and actual spot where reason was created).
 ///
 /// This is constructed via `.into()` on a `String`, `&'static str`, or
 /// `Box<str>`.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
rexmas: Would removal of these traits not be considered public API breaking? It would seem so to me since any client reliant on the methods of the removed traits would no longer work.

(Sorry for getting back to this so late, thank you for your patience 🙇)

target-san: These had to be implemented manually below - because Backtrace can't implement them, except Debug` which would be just too verbose. Please check lines 83+ and 165+.

PR #546, opened 2025-02-02
replace lazy_static with std::sync::LazyLock

serprex:


matthew-russo:
Thanks for the contribution.

We'll need to hold off for a few more Rust releases. Proptest's currently msrv policy is no more than current - 7:

```
# MSRV

The current MSRV of this crate is 1.64. The MSRV is guaranteed to not exceed <current stable release> - 7, though in practice it may be lower than this - your mileage may vary. If we change this policy in a backwards incompatible way (e.g. changing it to <current stable release> - 1), this constitutes a breaking change, and would be a major version bump (e.g. 1.1 -> 2.0).
```

https://github.com/proptest-rs/proptest?tab=readme-ov-file#msrv

I'll check back in a few months when 1.87 releases.

rexmas:
@serprex we're well past 1.87 if you would like to update this branch and our MSRV we can merge it

rexmas (approved):
👍

PR #549, opened 2025-02-20
Add PROPTEST_RNG_SEED to take initial TestRunner seed from environment

### LINKED ISSUE #548: Setting the seed globally from the environment?
I introduces some `proptest`s into a code base that have dramatically slowed down the CI job we use with all of the instrumentation flags set for code coverage (TLDR, it's not a problem with the proptest library). My team wants me to be able to print the seed, or even better pass the seed of choice in from the environment using some randomness from the environment. Here the tests are passing, so IIUC there is nothing RE seed being written to file system.

I see that this is possible using the `TestRunner`, reading an environment variable, setting the option, etc but that all sounds more manual than I would like since we currently use the macros. Is there currently / is there appetite for setting a seed from environment via something like PROPTEST_SEED env var?

martyall:
closes #548 

Ideally I wouldn't have to define the `RngSeed` type and could have just used an optional. However, the fact that `parse_or_warn` takes a `&mut T` and not a `&mut Option<T>` made it difficult to keep common parsing logic. Maybe you have another suggestion?

### DIFF
--- a/proptest/src/test_runner/config.rs
+++ b/proptest/src/test_runner/config.rs
@@ -37,6 +37,7 @@ pub fn contextualize_config(mut result: Config) -> Config {
     const TIMEOUT: &str = "PROPTEST_TIMEOUT";
     const VERBOSE: &str = "PROPTEST_VERBOSE";
     const RNG_ALGORITHM: &str = "PROPTEST_RNG_ALGORITHM";
+    const RNG_SEED: &str = "PROPTEST_RNG_SEED";
     const DISABLE_FAILURE_PERSISTENCE: &str =
         "PROPTEST_DISABLE_FAILURE_PERSISTENCE";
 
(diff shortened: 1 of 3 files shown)


matthew-russo (approved):
Thanks! I don't mind the `RngSeed` type. I'd rather keep common parsing logic

matthew-russo:
Seems like build is failing for no_std:
https://github.com/proptest-rs/proptest/actions/runs/13430756164/job/37654228898?pr=549#step:10:33

martyall:
ah my bad, wasn't even considering it. Ran locally with that workflow's build command and is now building.

matthew-russo (approved):

martyall:
@matthew-russo seems like the doc test is failing for reasons that are unrelated?

rexmas (approved):
this looks reasonable, thank you!

matthew-russo:
I'm fine to merge as is, I'll fix the build failure later

PR #554, opened 2025-03-04
chore: assert collection in `sample::select` is not empty

thomaseizinger:
This will already fail further down the call-chain but asserting it here gives much better stracktraces, allowing users to debug much faster, where in their potentially very large and composed strategy the problem is.


rexmas (approved):
yup, nice little find. ty

matthew-russo (approved):

PR #555, opened 2025-03-20
Update version number of proptest in instructions

felix91gr:
Version number for Cargo.toml file was 1.0.0.
Change it to 1.6.0


felix91gr:
The old version number actually confused me for a while. Now that should not happen again :)

rexmas (approved):
👍

PR #558, opened 2025-04-07
Config : support providing full hex-encoded seeds via PROPTEST_RNG_SEED

### LINKED ISSUE #556: Add `PROPTEST_RNG_SEED` to the output file on failure
It would be great if the regression file included the `PROPTEST_RNG_SEED` on failure. This would allow us to reproduce a run directly from the command line without needing the regression file for debugging.  

@moodmosaic and I rely on `proptest` for one of our latest projects. We're currently using b9eb3e4 because it lets us specify a seed directly from the command line. Adding the seed to the output file on failure would make debugging much easier.  

Thanks!

matthew-russo:
fixes #556

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Unreleased
 
+### New Features
+
+- Added the ability to provide a full hex-encoded seed via PROPTEST_RNG_SEED,
+  allowing users to roundtrip a seed persisted after a failed run via the
+  command line.
+
 ## 1.6.0
 
 ### New Features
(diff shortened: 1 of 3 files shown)


moodmosaic (commented):
@matthew-russo, thank you for working on this (via https://github.com/proptest-rs/proptest/issues/556) 🙏 ❤️ 

Can we avoid requiring the `hex-` prefix for replay? The process is already  
a bit complex. If the replay file is there, just use it. If not, the failure  
message should mention how to replay with `hex-`. Also, there's  
`PROPTEST_RNG_SEED` to consider. Adding another format rule adds friction.

---

Better: auto-detect the seed format.  
Try to parse the seed as a hex string of the right length.  
If that works, use it as a hex seed.  
Otherwise, parse it as a u64.

This lets users copy-paste seeds from failure output as-is.  
Also works cleanly with `PROPTEST_RNG_SEED`.

PR #574, opened 2025-05-20
Fix URLs in proptest-derive error messages

Nicolapps:
This PR fixes the error messages emitted by proptest-derive. Before, the URLs had the following format:

> https://PATH/TO/foo#E0001

Not only the URL doesn’t exist, but also the fragment is in uppercase characters (E0123) instead of the lowercase string (e0123) that rustdoc generates.

Now, they use the correct format instead:

> https://proptest-rs.github.io/proptest/proptest-derive/errors.html#e0001

I added a unit test to check the output format of `mk_err_msg`. Sadly it’s not possible to test the URL format in the compile-fail tests because it looks like the support from [compiletest-rs](https://github.com/Manishearth/compiletest-rs) for multi-line error messages is broken.

### DIFF
--- a/proptest-derive/src/error.rs
+++ b/proptest-derive/src/error.rs
@@ -206,15 +206,11 @@ impl Context {
 /// to the given `$message`.
 macro_rules! mk_err_msg {
     ($code: ident, $msg: expr) => {
-        concat!(
-            "[proptest_derive, ",
+        format!(
+            "[proptest_derive, {}] during #[derive(Arbitrary)]:\n{} Please see: https://proptest-rs.github.io/proptest/proptest-derive/errors.html#{} for more information.",
             stringify!($code),
-            "]",
-            " during #[derive(Arbitrary)]:\n",
             $msg,
-            " Please see: https://PATH/TO/foo#",
-            stringify!($code),
-            " for more information."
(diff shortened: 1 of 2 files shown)


proptest-derive/src/error.rs
```diff
@@ -229,7 +225,7 @@ macro_rules! fatal {
     ($error: ident ($($arg: ident: $arg_ty: ty),*), $code: ident,
      $msg: expr, $($fmt: tt)+) => {
         pub fn $error<T>(ctx: Ctx, $($arg: $arg_ty),*) -> DeriveResult<T> {
-            ctx.fatal(format!(mk_err_msg!($code, $msg), $($fmt)+))
+            ctx.fatal(mk_err_msg!($code, format!($msg, $($fmt)+)))
```
Nicolapps, line 228: I’m making sure only `$msg` is used as the format string of `format!` because `to_lowercase` isn’t `const`.

rexmas (approved):
👏 nice find

rexmas:
ci is fixed on main now so this should pass with a rebase.

Nicolapps:
@rexmas Thanks! I rebased and the CI passes now

PR #576, opened 2025-05-27
Rand 0.9

rexmas:
This PR updates rand to v0.9 based off the work of https://github.com/proptest-rs/proptest/pull/552 and https://github.com/proptest-rs/proptest/pull/562

PR #577, opened 2025-05-28
Upgrade `rand` to 0.9.1

### LINKED ISSUE #553: Update proptest to depend on rand 0.9
Rand 0.9 was released back in January. The rollout is pretty disruptive, since a crate dependency tree with both 0.9 and previous versions will fail to compile with some not-very user friendly errors about trait bounds not being satisfied: example

It would be a lovely thing if the propest crate would update its and dependency and enable us to move forward over this bump.

There have been some PRs [swirling](https://github.com/proptest-rs/proptest/pull/552) [around](https://github.com/proptest-rs/proptest/pull/550) but unfortunately the rand crate has removed it's implementations of `Distribution<usize>` and `Distribution<isize>`, which means that the code changes needed to support the update is not simply a mechanistic update to match the new api naming.

I'm assuming that rand will hold steady in its decision to remove the two impls, and I would also assume that this project is disinterested in removing existing functionality for those types. To make matters worse, a solution that combines the upstream StandardUniform distributions but attempts to add new local impls for isize/usize through extending the Rng trait would up hitting `E0119: conflicting implementations of trait" where the compiler helpfully tells us that `upstream crates may add a new impl of trait `rand::distr::Distribution<usize>` for type `rand::distr::StandardUniform` in future versions`

From where I'm standing, it seems the most straightforward solution to this is to build some small `random()` wrapper that would explicitly delegate it's functionality to `Rng::random()` for the types it handles and implement a `target_pointer_with` switched version of the usize/isize `random()`, but I'm open to the possibility that there might be some other way.

criccomini:
I took a crack at upgrading the `rand` dependency to 0.9.1 (latest). I also upgraded `rand_*` and added a `rand_core` dependency. I went through the [updating notes](https://rust-random.github.io/book/update-0.9.html) and made the appropriate changes.

My motivation for this PR is that `slatedb` uses `ulid`, `uuid`, and `proptest`. While updating `slatedb` to support seedable and portable RNGs for deterministic simulation testing, I discovered that the recent versions of `uuid` and `ulid` all use `^0.9.0`. This caused a lot of issues; ultimately, I had to drop all library versions way down to maintain compatibility with proptest, which requires ^0.8.0. Cargo seems to enforce compatibility across pre 1.0 minor versions, so it was really throwing a fit.

The upgrade was fairly straightforward except for the changes to `isize` and `usize`. As described in #553, `rand` dropped those because they're not portable. It does not seem they're likely to add them back (https://github.com/rust-random/rand/pull/1598).

I'm assuming that `proptest` wants to keep support for `isize` and `usize` distributions and random numbers. I tried several strategies to maintain these features without the underlying support in `rand`. Ultimately, I settled on adding `next_isize` and `next_usize` to `TestRunner`. I subsequently updated the macros in `num.rs` to call `next_isize` and `next_usize` when `$typ` is an `isize` or `usize`. This required implementing strategies. The results are a little clunky, but I feel they strike the best balance between maintainability and ease of implementation.

If others believe there's a better approach, I'm happy to listen.

Fixes #553

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,10 @@ prettyplease = "0.2"
 proc-macro2 = "1.0"
 proptest-macro = { version = "0.1", path = "proptest-macro" }
 quote = "1.0"
-rand = { version = "0.8", default-features = false }
-rand_chacha = { version = "0.3", default-features = false }
-rand_xorshift = "0.3"
+rand = { version = "0.9.1", default-features = false, features = ["os_rng"] }
+rand_core = { version = "0.9.3", default-features = false }
+rand_chacha = { version = "0.9.0", default-features = false }
+rand_xorshift = "0.4.0"
 regex = "1.0"
 regex-syntax = "0.8"
 rusty-fork = { version = "0.3.0", default-features = false }
--- a/proptest/Cargo.toml
+++ b/proptest/Cargo.toml
@@ -72,6 +72,7 @@ regex-syntax = { workspace = true, optional = true }
 bit-set = { workspace = true, optional = true }
 bit-vec = { workspace = true, optional = true }
 rand = { workspace = true, features = ["alloc"] }
+rand_core = { workspace = true }
 rand_xorshift = { workspace = true }
 rand_chacha = { workspace = true }
 rusty-fork = { workspace = true, optional = true }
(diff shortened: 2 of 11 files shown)


rexmas:
thank you for putting in the work, however the following rand upgrade has been merged https://github.com/proptest-rs/proptest/pull/576 . if you believe there's anything else in your pr that you think is important please rebase on top of main and open a new pr, thanks!

PR #589, opened 2025-07-31
Add `ProptestResultExt` with `prop_assume_ok` helper

bschoenmaeckers:
I found myself using `prop_assume!(result.is_ok())` a lot before unwrapping some `Result` value that I use while testing. So to make life easier I come to the following api addition. 

This allows to write test like the following.

```rust
use jiff::civil::Date;
use proptest::prelude::*;

proptest! {
  #[test]
  fn test_some_date(
    year in 1i16..=9999i16,
    month in 1i8..=12i8,
    day in 1i8..=31i8
  ) {
    // Reject invalid dates
    let date = Date::new(year, month, day).prop_assume_ok()?;
    // ...rest of test...
  }
}
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -48,6 +48,8 @@ jobs:
         run: cd proptest && cargo test --verbose
       - name: Run macro tests
         run: cd proptest-macro && cargo test --verbose
+      - name: Run macro integration tests
+        run: cargo test --verbose --test attr_macro --features attr-macro
       - name: Build coverage no-default-features
         if: ${{ matrix.build == 'stable' }}
         env:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ message-io = { version = "0.19.0", default-features = false, features = [
 num-traits = { version = "0.2.15", default-features = false }
 prettyplease = "0.2"
 proc-macro2 = "1.0"
-proptest-macro = { version = "0.2", path = "proptest-macro" }
+proptest-macro = { version = "0.3.1", path = "proptest-macro" }
 quote = "1.0"
 rand = { version = "0.9", default-features = false }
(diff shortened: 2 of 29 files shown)


proptest/src/test_runner/errors.rs
```diff
@@ -132,3 +132,48 @@ impl<T: fmt::Debug> ::std::error::Error for TestError<T> {
         }
     }
 }
+
+/// Extension trait for `Result<T, E>` to provide additional functionality
+/// specifically for prop test cases.
+pub trait ResultExt<T, E> {
```
rexmas: I think this is a nice ergonomic addition, but `ResultExt` is pretty common. can we name this `ProptestResultExt` instead?

also seems like a good case for a sealed trait
e.g. from [blog post](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/)
```rust
mod private {
    pub trait Sealed {}
}

pub trait SealedTrait : private::Sealed {
    fn method(&self);
}
```

proptest/src/test_runner/errors.rs
```diff
@@ -132,3 +132,48 @@ impl<T: fmt::Debug> ::std::error::Error for TestError<T> {
...
+    ///
+    /// [`prop_assume!`]: crate::prop_assume
+    fn assume_ok(self) -> Result<T, TestCaseError>
+    where
+        E: fmt::Debug;
+}
+
+impl<T, E> ResultExt<T, E> for Result<T, E> {
+    #[track_caller]
+    fn assume_ok(self) -> Result<T, TestCaseError>
```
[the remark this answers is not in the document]
bschoenmaeckers: Sounds good to me. Changed.

proptest/src/test_runner/errors.rs
bschoenmaeckers: Added.

PR #594, opened 2025-09-04
update attr macro to use argument names where trivial

### LINKED ISSUE #592: property_test struct generation doesn't respect field names
```rust
#[property_test(config = ProptestConfig { verbose: 2, ..ProptestConfig::default() })]
fn my_test(
    foo: bool,
    #[strategy = arbitrary_bar()] bar: u64,
) {
    let _ = foo;
    let _ = bar;
}
```

results in prints such as
```
proptest: Next test input: MyTestArgs = MyTestArgs { field0: false, field1: 12305979233754266495 }
proptest: Test case passed
```
and I prefer `MyTestArgs = MyTestArgs { foo: false, bar: 12305979233754266495 }` which makes the test case more clear.

cameron1024:
Closes #592

Note for reviewers: 1750 lines of diff comes from the inclusion of a new snapshot file. It's otherwise quite a small PR.

The generated struct will now use the name of an argument *if* the pattern of the argument is an identifier. For example, given this failing test:
```rust
#[property_test]
fn test_name(x: i32, #[strategy = "[0-9]*"] s: String) {
    assert_eq!(x.to_string(), s);
}
```
We now get an error like:
```
Test failed: assertion `left == right` failed
  left: "0"
 right: "1".
minimal failing input: TestNameArgs = TestNameArgs {
    x: 0,
    s: "1",
}
```

There are two possible improvements (that I can think of off the top of my head right now) here:

- we could extend this to cover *any pattern* that binds only a
single parameter, for example, we could capture this argument as `s`:
```rust
#[property_test]
fn foo((Ok(s) | Err(s)): Result<i32, i32>) {}
```
- we could do some sort of name mangling for more common pattern types, for
  example `(a, b): (i32, i32)` could correspond to a struct field
  `tuple_of_a_b: (i32, i32)`.

Of these, the second option is a lot easier to implement, though it could get nasty with nested patterns (e.g. `tuple_of_tuple_of_tuple_of...`).

For now I'm going to get this minimal implementation in though.

### DIFF
--- a/proptest-macro/src/property_test/codegen/snapshots/proptest_macro__property_test__codegen__snapshot_tests__arg_ident_and_pattern.snap
+++ b/proptest-macro/src/property_test/codegen/snapshots/proptest_macro__property_test__codegen__snapshot_tests__arg_ident_and_pattern.snap
@@ -1479,13 +1479,6 @@ TokenStream [
                                         sym: a,
                                     },
                                     Punct {
-                                        char: ':',
-                                        spacing: Alone,
-                                    },
-                                    Ident {
-                                        sym: a,
-                                    },
-                                    Punct {
                                         char: ',',
                                         spacing: Alone,
                                     },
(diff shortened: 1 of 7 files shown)


rexmas (approved):
awesome!

PR #595, opened 2025-09-12
Add `ReferenceStateMachine` arg to `SystemUnderTest::teardown`

ebegumisa:
Regarding [state machine tests][1], I have frequently found it useful to be able to check certain properties only after all transitions have been applied (in addition to other properties checked after individual transitions). 

To this end, this PR makes the reference state machine available to compare against during SUT teardown.

[1]: https://proptest-rs.github.io/proptest/proptest/state-machine.html


rexmas (approved):
👍

PR #596, opened 2025-09-14
fix for 32 bit processors

alexanderkjall:
based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/proptest/debian/patches/fix-arithmetic-overflow.patch?ref_type=heads


rexmas:
this is great, thanks

alexanderkjall:
No problem, thanks for merging :)

I now noticed that I missed to update the comment right above the code,

PR #609, opened 2025-11-06
feat(proptest-state-machine): allow closures in `Sequential::new`

thomaseizinger:
Currently, `Sequential::new` only takes function pointers for the various hooks of the state machine test. This is limiting for advanced use cases of this crate where `Sequential` is constructed outside of the `ReferenceStateMachine` trait. We do that in https://github.com/firezone/firezone/blob/main/rust/connlib/tunnel/src/tests.rs to further customize the way this test is run.

I'd like to extend this to accept closures instead to allow passing additional parameters to the `apply` and `initial_state` functions that are captured within the closure itself.

You can pass a function pointer wherever you can pass a pointer. As such, I think this is a backwards-compatible change.

### DIFF
--- a/proptest-state-machine/src/strategy.rs
+++ b/proptest-state-machine/src/strategy.rs
@@ -42,7 +42,7 @@ use proptest::test_runner::TestRunner;
 /// The reference state machine generation runs before the generated transitions
 /// are attempted to be executed against the SUT (the concrete state machine)
 /// as defined by [`crate::StateMachineTest`].
-pub trait ReferenceStateMachine {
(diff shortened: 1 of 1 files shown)


thomaseizinger:
CI failures seem to be unrelated, looks like an MSRV bump in a minor version of a dependency?

tzemanovic:
> CI failures seem to be unrelated, looks like an MSRV bump in a minor version of a dependency?

ah yes, it's a dev-dependency only - I'll take care of it in a different PR

thomaseizinger:
Thank you!

PR #612, opened 2025-11-15
Fix state-machine CI pinned test deps versions

tzemanovic:
- Some dev-dependencies of state-machine have been updated to MSRV 1.83, but we're on 1.82, causing a failure in pinned CI test (e.g. https://github.com/proptest-rs/proptest/actions/runs/19123262762/job/55424850296). Locked the versions of the said dependencies
- Added changelog for #609


rexmas (approved):
Nice

PR #618, opened 2025-12-18
add license files to proptest-state-machine

martinvonz:
The other three crates already have license files. I'm guessing it's just an oversight that proptest-state-machine did not.


rexmas (approved):
thank you

PR #635, opened 2026-03-06
`proptest_path` in attr macro

cameron1024:
Adds an argument to the attribute macro `proptest_path = "some::path::to::proptest"`, which changes where the macro assumes it can find the proptest crate. I.e., if you provide `proptest_path = "some::path::to::proptest`, the macro will emit:
```rust
impl some::path::to::proptest::prelude::Arbitrary for ... {}
```

## Motivation

At Zed, we have a [custom test macro](https://github.com/zed-industries/zed/blob/2457e27437b355f030793f2085cda51bbc39b642/crates/gpui_macros/src/gpui_macros.rs#L186) that injects some special arguments, and provides a limited form of property testing. I am improving this macro by using proptest instead of uniformly sampling from a distribution, but the current macro assumes that `::proptest` will always be accessible. Adding this parameter allows the `gpui::test` macro to expand to `#[gpui::proptest::property_test(proptest_path = "gpui::proptest") fn ...`, which removes the need to add `proptest` as a dependency in every crate that wants to write a test.


proptest-macro/src/property_test/options.rs
```diff
@@ -11,6 +10,19 @@ pub(super) struct Options {
...
     pub config: Option<Expr>,
+    pub proptest_path: Option<LitStr>,
+}
+
+impl Options {
+    pub fn true_proptest_path(&self) -> TokenStream {
+        match &self.proptest_path {
+            None => quote! { ::proptest },
+            Some(s) => {
+                let value: Path = parse_str(&s.value()).unwrap();
```
rexmas, line 21: unwrap might be a bit confusing without an error message. also it doesn't look like we parse for path syntax, there might be a way to do that with the `Path` type though https://docs.rs/syn/latest/syn/struct.Path.html and add a test against bogus parses. thoughts?

cameron1024: Good thinking, ended up changing it to take just a path, without the string-literal wrapping. For the parsing, we'll get an error message from `syn` if it fails. We also get autocomplete "for free", so feels like the better choice

rexmas (approved):
looks great 👍 will fix ci in a follow up

Vanuan:
Any chance new crate could be released to remove git pin from zed/gpui?

PR #645, opened 2026-04-28
fix nightly ambigous SIGN_MASK and MANTISSA_MASK

tzemanovic:
switched to use a fully qualified name with these 2 associated constants to avoid ambiguity with ones from nightly `float_masks` feature


rexmas (approved):
tyty

PR #647, opened 2026-05-13
fix: handle single-point inclusive float range without panic

### LINKED ISSUE #479: RangeInclusive of floats containing one element panics
Following code panics:

```rust
#[test]
fn test(x in 0.0..=0.0) {

}
```

Even though `0` is in `0..=0` interval. This does not happen with the integers. Following is a workaround on unstable:

```rust
#[test]
fn test(x in 0.0..0.0_f32.next_up()) {

}

SAY-5:
Sampling from a single-point inclusive float range like `0.0..=0.0` panicked in `split_interval` because the assert required `high - low > 0.`. Skip the interval split when `low == high` in `new_inclusive` and return that value directly. Closes #479.

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - The minimum supported Rust version has been increased to 1.86.0.
 
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


tzemanovic (approved):
lgtm, thanks!

PR #649, opened 2026-06-21
[fix] replace unsound static mut access with OnceLock

### LINKED ISSUE #650: Data Race on static mut DEFAULT_HOOK During Panic Hook Initialization
See #648 for description

matthew-russo:
Fixes #650


rexmas (approved):
yup, good find!

PR #652, opened 2026-06-21
[fix] assert that rdrand is supported before making unsafe calls that expect its existence

### LINKED ISSUE #651: Soundness Violation: Unverified Target Feature Execution in Safe API hardware_rng

matthew-russo:
fixes #651


rexmas (approved):
makes sense, thanks! 👍

Issue #440, opened 2024-04-06
Recommendation to check in proptest regressions seems off


vlovich:
This is purely subjective of course, but imagine you do this & your code has 250 regression test cases. You probably won't notice meaning every time you run proptest it's only testing the regressions that have been fixed instead of testing new iterations.

In practice (from JS quickcheck which is where I had more experience), I found myself adding explicit test cases when encountering a bug. That way I can explicitly document the specific conditions that leads to the bug and have an explicitly documented test case for the regression while leaving the property testing to continue randomly exploring potential other bugs.

matthew-russo:
I see a few related but different topics here:

> explicit test cases

we're looking at adding first-class support for these so you can share test code between arbitrary cases and well-known edge cases. i think longer term this will be an ergonomic approach to dealing with this and you can disable persistence and just list our your explicit cases

> validity of a seed against changing properties

i don't have a perfect answer for this one. its definitely an issue, especially for areas that have some churn on the underlying data that gets generated. i don't think this makes it fundamentally useless but there are definitely cases where it becomes borderline useless against its original goal. the few alternative/complementary approaches I can think of off the top of my head aren't foolproof either:
- we could add some type of fingerprint/signature/hash of the data generated via Arbitrary and persist this alongside the seed and when a mismatch is found between the hash on-disk and the hash generated in memory, we could print a warning informing the user that their seed is no longer generating the same data. this might be pretty invasive in terms of implementation -- haven't thought about it too deeply, and it still leaves effort on the end-user to go clean up stale seed
- we could attempt to serialize the actual values generated, rather than/in addition to the original seed and then re-use the values verbatim. this also seems a bit invasive as we'd need to be able to serialize/deserialize generated values. it would also lead to variable sizes of the persisted file and since the generated data is... Arbitrary, i'm not a big fan of this one

> persisted cases decrease from the number of fuzzed cases

this one isn't quite right. we don't count the persisted cases in our total number of runs so if you say 10 cases and you have 5 persisted seeds, we'll run the 5 persisted seeds and then our count is still at 0 and then run the 10 cases with new generated data. now this comes with the downside of increased execution time but if we assume the fidelity of the persisted seeds is good, then it's effectively the same run time as if you had manually written out the 5 explicit cases

> explicit cases vs seed, "misusing proptest", etc

[...]

PR #653, opened 2026-06-24
feat(state-machine): persist shrunk transition sequence, not just the seed

### LINKED ISSUE #564: Add a way to save and load fully-shrunk test cases
I usually want to examine failing test cases in my debugger, ideally skipping irrelevant steps from an un-shrunk version of the case.

Everything seems pretty well set up to do that with proptest, with the persistence system and PROPTEST_CASES=0. But IIUC, because all that's persisted is the seed of the failing case, proptest can't avoid repeating the shrinking process when replaying from the file.

I can always copy the handy printing of the failing case to a separate test, but I only rarely want to add stress test failures as permanent regression tests, so it's just a bit of awkward busywork. Or I can figure out a strategy for skipping the shrinking steps and finding the start of the fully-minimized case's execution in my debugger. But it'd be nice to skip all that with some opt-in system for saving and loading fully-shrunk outputs.

There's some discussion of reasons *not* to do this at https://proptest-rs.github.io/proptest/proptest/failure-persistence.html, but it seems to me that the seed system isn't different from the value system in the main way that counts: accurately replaying the failing execution. And as I think https://github.com/proptest-rs/proptest/issues/440 is saying, divergence isn't the only reason why I wouldn't want to persist the file beyond replaying the most recent execution of a test against a known version of the code. So the "persisted case is at least guaranteed valid" point doesn't seem to buy me much. The auto-regression-test use case of checking the persistence file into git is neat, but *that's* where I would actually want to check the printed case in as code, to make it less likely that my regression testing quietly goes out of date.

Also, regardless, I'd personally happily deal with surprising outcomes in the face of code changes if it meant that I could set up a comfy debugging experience when stress testing code that I'm very actively working on.

nightscape:
# Persist the shrunk transition sequence, not just the seed (proptest-state-machine)

Closes #564. Related context: #440.

## Problem

Proptest's built-in failure persistence stores the **RNG seed** of the failing
`TestRunner`. For state-machine tests that has two costs:

1. **Replay re-shrinks.** Replaying a seed regenerates the case from the
   strategy and then re-runs the whole shrinking process before the minimal
   counter-example is reached.
2. **Seeds are brittle.** As soon as the reference model or the transition
   strategy changes, the same seed no longer maps to the same case, so the
   stored regression silently stops pinning the bug it was meant to pin.

Issue #564 asks for the ability to save and load the *fully-shrunk* failing
case directly. Unlike a generic proptest `Value`, a state-machine failing case
is a concrete, human-readable value — `(initial_state, Vec<Transition>)` — that
can usually implement `serde::Serialize`. That makes the state-machine crate the
natural home for this (proptest core stays seed-based and untouched apart from a
one-line env-var namespace reservation).

## What this adds

A new **opt-in `persistence` feature** (pulls in `serde` + `serde_json`).

The model mirrors how proptest treats seed regressions: persisted cases are
**replayed on every run**, and freshly generated cases run on top to keep
finding new failures. Distinct failures **accumulate** into a regression file
stored as **JSON Lines** (one case per line, like proptest's line-oriented
seed-regression file — clean diffs and union-friendly merges in source control),
while shrunk versions of the *same* failure collapse to one minimal case.

- `prop_state_machine_persisted!` — macro counterpart of `prop_state_machine!`.
  It expands to a test that (1) calls `replay_persisted_regressions` once per
  run, then (2) runs the normal generation loop with capture-on-failure.
- `StateMachineTest::replay_persisted_regressions(config)` — loads the persisted
  case (if any) and re-runs it via `test_sequential`. A still-failing regression
  panics here, attributed to the stored case rather than to a generated one.
- `StateMachineTest::test_sequential_persisted(...)` — `test_sequential` plus
  capture-on-failure.
- `persistence::PersistedCase<S, T>` — the on-disk case shape
  (`{ initial_state, transitions }`); the file is one such object per line
  (JSON Lines) under a `#` comment header.
- `persistence::store` — the `StateMachineTest`-agnostic half: JSON Lines I/O,
  the accumulate/de-dup merge, the per-run marker, and the panic `CaptureGuard`,
  generic over any `Serialize + DeserializeOwned` case type. The trait/macro
  glue is a thin layer on top. Factored out so it's reviewable in isolation and
  reusable by projects with a richer fixture type; happy to narrow its
  visibility if you'd prefer a smaller public surface.
- `persistence::CaptureGuard` — an RAII guard that serializes the case eagerly
  and, on unwind, merges it into the regression set. Within a run, proptest
  re-runs the body for every shrink candidate; successive writes replace this
  run's previous entry (collapsing to the minimal). Across runs, a distinct
  minimal is appended; exact duplicates are de-duplicated.
- `PROPTEST_STATE_MACHINE_PERSIST_DIR` env var overrides the persistence
  directory (default `proptest-regressions/state-machine`). `PROPTEST_CASES=0`
  replays the regression without generating new cases.
- proptest core: the `PROPTEST_STATE_MACHINE_*` namespace is reserved so the
  config env scanner doesn't print `Ignoring unknown env-var`.

The reference `State` and `Transition` must implement
`serde::Serialize + serde::de::DeserializeOwned` to use the persisted path; the
base traits are unchanged, so existing tests and default builds are completely
unaffected.

## Usage

```rust
prop_state_machine_persisted! {
    #[test]
    fn my_test(sequential 1..20 => MyStateMachine);
}
```

On failure the shrunk sequence is written to
`proptest-regressions/state-machine/<type>.jsonl` and committed like a normal
regression file. Every subsequent run replays the stored cases first, then
generates new cases. To replay only the regressions (no new generation):

```sh
PROPTEST_CASES=0 cargo test my_test
```

## Tests

`proptest-state-machine/tests/persistence.rs`:

- a deliberately buggy counter machine (SUT saturates at 3) fails; the persisted
  file contains the **minimal** shrunk sequence (exactly four increments), not
  an unshrunk one;
- `replay_persisted_regressions` reproduces the failure from that file;
- a passing machine writes nothing;
- a correct machine driven through `prop_state_machine_persisted!` exercises the
  macro (replay no-op + generation) end to end;
- a machine with two selectable bugs accumulates **two distinct** minimal cases
  across runs, and re-discovering one already stored does not duplicate it
  (plus a focused unit test of the set-merge logic).

## Notes / open questions for maintainers

- **Collapsing "shrunk versions of each other".** Distinct minimals are kept by
  exact-equality de-dup, not by subsequence subsumption. Subsumption would be
  unsound here: two genuinely different bugs can be subsequence-related, so
  collapsing by subsequence could silently drop a distinct regression. Instead,
  the replay-before-generate flow already prevents accumulating two shrinks of
  the *same* live bug (its regression fails replay before generation can run),
  and within-run shrinking collapses to one minimal. If you'd still like
  same-bug-across-strategy-change collapsing, we'd need a per-failure identity
  signal (e.g. the panic message) — happy to explore.
- **Why `State: Serialize` and no bespoke abstraction?** Persistence is a single,
  uniform serde requirement on both `State` and `Transition` — no special-casing
  of one over the other. A state that is large or not directly serializable can
  persist a compact, reconstructible form via `#[serde(into = "…", from = "…")]`
  (documented in the module docs), so the serde bound isn't a hard ceiling. I
  deliberately avoided a dedicated state-codec trait: it would be asymmetric with
  `Transition` and is redundant with serde's own proxy support for any state you
  own. If foreign (orphan-rule-blocked) states create real demand later, a codec
  can be added additively.
- The capture/replay relies on the in-process unwind path; it is not wired
  through the `fork`/`timeout` out-of-process runners. Happy to gate or document
  that explicitly.
- JSON via `serde_json` was chosen for human-readability (these files are meant
  to be inspected and checked in). If you'd prefer a different format or to make
  the serializer pluggable, easy to change.
- Naming of the trait method / macro / env var is open to bikeshedding.

### DIFF
--- a/proptest-state-machine/src/test_runner.rs
+++ b/proptest-state-machine/src/test_runner.rs
@@ -135,24 +135,14 @@ pub trait StateMachineTest {
         Self::teardown(concrete_state, ref_state)
     }
 
-    /// Like [`test_sequential`](Self::test_sequential), but on failure persists
-    /// the (shrunk) `(initial_state, transitions)` to disk so it can be
-    /// replayed later without regenerating from a seed or re-running the
-    /// shrinking process. See the [`persistence`](crate::persistence) module.
-    ///
-    /// If the case fails, the *last* failing case seen during shrinking — i.e.
-    /// the minimal one — is written to [`persistence::default_persist_path`]
-    /// (the directory is overridable via [`persistence::PERSIST_DIR_ENV`]). A
-    /// passing case writes nothing. The persisted case is replayed by
-    /// [`replay_persisted_regressions`](Self::replay_persisted_regressions);
-    /// [`prop_state_machine_persisted`](crate::prop_state_machine_persisted)
-    /// wires both together.
+    /// Like [`test_sequential`](Self::test_sequential), but a failing case is
+    /// merged into the regression set at `path` — build it with
+    /// [`persist_path!`](crate::persist_path). The minimal shrunk case is the
+    /// one that survives; a passing case writes nothing.
     ///
     /// Requires the reference `State` and `Transition` to implement
-    /// [`serde::Serialize`] + [`serde::de::DeserializeOwned`].
-    ///
-    /// [`persistence::PERSIST_DIR_ENV`]: crate::persistence::PERSIST_DIR_ENV
-    /// [`persistence::default_persist_path`]: crate::persistence::default_persist_path
+    /// [`serde::Serialize`] + [`serde::de::DeserializeOwned`]. Does nothing
+    /// beyond `test_sequential` when failure persistence is disabled.
     #[cfg(feature = "persistence")]
     fn test_sequential_persisted(
         config: Config,
@@ -173,31 +163,22 @@ pub trait StateMachineTest {
             return;
         }
         crate::persistence::assert_same_process(&config);
-        // Arm a guard that merges the case on unwind. proptest re-runs this body
-        // for every shrink candidate, so the final (minimal) failing case is
-        // the last write within the run and wins.
         let case = crate::persistence::PersistedCase {
             initial_state: ref_state.clone(),
             transitions: transitions.clone(),
         };
-        let guard =
-            crate::persistence::store::CaptureGuard::arm(path, &case);
+        let guard = crate::persistence::store::CaptureGuard::arm(path, &case);
         Self::test_sequential(config, ref_state, transitions, seen_counter);
-        // Passing case: do not persist.
         guard.disarm();
     }
 
-    /// Replay every persisted regression for this test, in order, by running
-    /// each through [`test_sequential`](Self::test_sequential). Returns the
-    /// number of cases replayed.
+    /// Replay every regression stored at `path`, in order, through
+    /// [`test_sequential`](Self::test_sequential). Returns how many ran.
     ///
-    /// This is meant to run on *every* test invocation — before generating new
-    /// cases — mirroring how proptest replays its seed regressions. A persisted
-    /// case that still fails panics here, with the failure attributed to the
-    /// stored regression rather than to a freshly generated case.
-    /// [`prop_state_machine_persisted`](crate::prop_state_machine_persisted)
-    /// calls this for you. It also resets the per-run accumulation marker, so a
-    /// failure generated afterwards starts a fresh entry in the regression set.
+    /// Call this before generating new cases; a stored case that still fails
+    /// panics here, so the failure is reported as the regression it is. Also
+    /// resets the per-run accumulation marker, so a failure generated
+    /// afterwards starts a fresh entry in the set.
     ///
     /// Requires the reference `State` and `Transition` to implement
     /// [`serde::Serialize`] + [`serde::de::DeserializeOwned`].
--- a/proptest-state-machine/src/test_runner.rs
+++ b/proptest-state-machine/src/test_runner.rs
@@ -134,6 +134,109 @@ pub trait StateMachineTest {
 
         Self::teardown(concrete_state, ref_state)
     }
+
+    /// Like [`test_sequential`](Self::test_sequential), but on failure persists
+    /// the (shrunk) `(initial_state, transitions)` to disk so it can be
+    /// replayed later without regenerating from a seed or re-running the
+    /// shrinking process. See the [`persistence`](crate::persistence) module.
+    ///
+    /// If the case fails, the *last* failing case seen during shrinking — i.e.
+    /// the minimal one — is written to [`persistence::default_persist_path`]
+    /// (the directory is overridable via [`persistence::PERSIST_DIR_ENV`]). A
+    /// passing case writes nothing. The persisted case is replayed by
+    /// [`replay_persisted_regressions`](Self::replay_persisted_regressions);
+    /// [`prop_state_machine_persisted`](crate::prop_state_machine_persisted)
+    /// wires both together.
+    ///
+    /// Requires the reference `State` and `Transition` to implement
+    /// [`serde::Serialize`] + [`serde::de::DeserializeOwned`].
+    ///
+    /// [`persistence::PERSIST_DIR_ENV`]: crate::persistence::PERSIST_DIR_ENV
+    /// [`persistence::default_persist_path`]: crate::persistence::default_persist_path
+    #[cfg(feature = "persistence")]
+    fn test_sequential_persisted(
+        config: Config,
+        ref_state: <Self::Reference as ReferenceStateMachine>::State,
+        transitions: Vec<
+            <Self::Reference as ReferenceStateMachine>::Transition,
+        >,
+        seen_counter: Option<Arc<AtomicUsize>>,
+    ) where
+        <Self::Reference as ReferenceStateMachine>::State:
+            serde::Serialize + serde::de::DeserializeOwned,
+        <Self::Reference as ReferenceStateMachine>::Transition:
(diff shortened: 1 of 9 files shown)


proptest-state-machine/src/persistence.rs
```diff
@@ -0,0 +1,208 @@
...
+// except according to those terms.
+
+//! Persist the *shrunk transition sequence* of a failing state-machine test,
+//! rather than only the RNG seed.
+//!
+//! Proptest's built-in failure persistence stores the seed of the failing
+//! `TestRunner`. Replaying that seed regenerates the case from the strategy and
+//! then **re-runs the whole shrinking process** before the minimal counter-
+//! example is reached. It is also brittle: once the reference model or the
+//! transition strategy changes, the same seed no longer maps to the same case.
```
nightscape: This comment has transient value: As soon as this PR is merged, the statement is not true anymore.

proptest-state-machine/src/persistence.rs
```diff
@@ -0,0 +1,208 @@
...
+//!
+//! For state-machine tests the failing case is a concrete, human-readable
+//! value — `(initial_state, Vec<Transition>)` — that, unlike a generic proptest
+//! `Value`, can usually implement [`serde::Serialize`]. This module persists
+//! that value directly, so a later run can replay the exact minimal sequence
+//! with no regeneration and no re-shrinking.
+//!
+//! See <https://github.com/proptest-rs/proptest/issues/564>.
```
nightscape: Not sure if this is good practice to link to a GitHub issue.
Imo, the code comments should be as self-sufficient as possible.
What useful information does the issue contain?
Please add it here if required.

proptest-state-machine/src/test_runner.rs
```diff
@@ -134,6 +134,80 @@ pub trait StateMachineTest {
...
+
+    /// Like [`test_sequential`](Self::test_sequential), but on failure persists
+    /// the (shrunk) `(initial_state, transitions)` to disk so it can be
+    /// replayed later without regenerating from a seed or re-running the
+    /// shrinking process. See the [`persistence`](crate::persistence) module.
+    ///
+    /// Behavior:
+    /// - If [`persistence::REPLAY_ENV`] is set, the generated `ref_state` /
+    ///   `transitions` are ignored and the stored case at that path is run
+    ///   instead (run your test with `PROPTEST_CASES=0` to replay only it).
```
nightscape: Is this the same behaviour as for the RNG-persisted regressions?
Imo, the more useful approach would be if the persisted regressions were one of the tried scenarios, not an "either persisted, or generated" approach.
We basically want to run persisted regressions every time, and generated ones on top to find unknown unknowns.

proptest-state-machine/src/persistence.rs
nightscape: Good catch — reworded so it describes the durable seed-based vs case-based trade-off rather than framing it as "what proptest does today". The seed path stays in the tree, so I scoped the statement to the *seed-based* persistence explicitly instead of implying it covers all of proptest. (9f26f58)

proptest-state-machine/src/persistence.rs
nightscape: Agreed — dropped the `See <issue>` link from the source. The motivation it pointed at (re-shrinking on replay + seed brittleness) is already spelled out in the module doc above, so the comment is self-sufficient now. I kept the issue reference in the CHANGELOG and PR body, which feels like the right place for provenance. (9f26f58)

proptest-state-machine/src/test_runner.rs
nightscape: Great point — the either/or was weaker than proptest's own behaviour. Reworked it: persisted regressions now replay on **every** run before generation, like seed regressions. `prop_state_machine_persisted!` calls the new `replay_persisted_regressions` once per run, then runs the normal generation loop on top — so you get "known regressions always + fresh fuzzing for unknown unknowns". `PROPTEST_CASES=0` replays just the regression. The `PROPTEST_STATE_MACHINE_REPLAY` env var is gone.

One open question (also in the PR notes): proptest's seed file *accumulates* multiple distinct regressions, whereas this currently keeps a single shrunk case per test (latest failure wins). Accumulating multiple (e.g. one file per content hash) is doable, but it fights the "last shrink-write wins" capture trick a bit, so I left v1 single-case. Want me to add multi-regression accumulation for full parity with seed regressions? (9f26f58)

nightscape: Done — implemented accumulation (058df5c). The persisted file is now a JSON **array** of cases, like proptest's seed-regression file: distinct failures accumulate across runs, while shrunk versions of the *same* failure collapse to one minimal (within-run shrink writes replace this run's previous entry, tracked by a marker reset each run in `replay_persisted_regressions`).

On the "shrunk versions of each other" point: I kept de-dup to **exact equality** rather than subsequence subsumption, because subsumption is unsound here — two genuinely different bugs can be subsequence-related, so collapsing by subsequence could silently drop a distinct regression. The replay-before-generate flow already prevents accumulating two shrinks of the *same* live bug (its regression fails replay before generation runs). New tests cover accumulation of two distinct minimals + no-duplicate on re-discovery.

nightscape: Switched the on-disk format to **JSON Lines** (81555cd) — one compact case per line under a `#` comment header, matching proptest's line-oriented seed-regression files. Adding a regression is now a one-line diff, and two branches each adding one merge as a clean union, which the pretty-printed array did not give. (`.json` → `.jsonl`; dropped the per-case `note` in favour of the header.)

PR #657, opened 2026-07-11
Fix panic in Arbitrary for Duration on nanosecond carry overflow

matthiasgoergens:
`any::<Duration>()` can panic when normalising its arbitrary nanosecond component would overflow the seconds counter.

Discard the carry only for affected inputs. A regression test exercises `(u64::MAX, u32::MAX)` through the actual strategy using a pass-through RNG.

### DIFF
--- a/proptest/CHANGELOG.md
+++ b/proptest/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


rexmas (approved):
nice find, technically this might add a small bias at the boundary but in the grand scheme of things seems negligible

PR #658, opened 2026-07-11
Choice-tape shrink engine: Conjecture-style generation recording, shrinking, and persistence

matthiasgoergens:
This PR adds an experimental choice-tape shrinking engine to proptest. Generation records typed choices and raw RNG reads; shrinking edits that record, replays the strategy, and keeps proposals that still reproduce the failure with a simpler tape.

The aim is to improve cases where independent `ValueTree` shrinking stalls: filtered values, related values that must move together, structured deletion, and floats whose simplest failing example is a round boundary. Existing strategies remain usable. Migrated strategies provide typed constraints and shrink targets, while other strategies participate through intercepted raw RNG draws with less specialised shrinking.

Collections use continuation choices and structural spans, allowing elements to be deleted independently. The sequential state-machine strategy uses the same representation and rechecks preconditions against the replayed state. The shrinker also includes per-choice minimisation, batched span deletion, redistribution between related integers, and joint lowering of duplicate values.

Failing tapes are persisted in the versioned `ct1` format. This reproduces the recorded choices rather than reconstructing them from an RNG seed, but it remains positional implementation data rather than a stable cross-version format. Custom persistence implementations need the newer persistence hook; legacy seed-only hooks cannot store tape failures.

This design follows Hypothesis's Conjecture engine. Hegel, developed by Hypothesis authors, is an earlier Rust implementation of the same architecture and likewise records typed choices with their constraints, including per-integer `shrink_towards` values. The contribution here is adapting the Hypothesis model to proptest's existing `Strategy`, `ValueTree`, `TestRunner`, RNG, and persistence interfaces rather than introducing the typed-choice representation itself.

The tape engine remains experimental and opt-in through `Config::shrink_engine` or `PROPTEST_SHRINK_ENGINE=tape`; `ValueTree` remains the default. Numeric edge-case generation changes are independent of the selected shrink engine and are called out separately in the changelog.

### Testing

- `cargo test -p proptest`
- `cargo test -p proptest-state-machine`
- `PROPTEST_SHRINK_ENGINE=tape cargo test -p proptest-state-machine`
- `cargo clippy -p proptest --lib`

The design and remaining limitations are documented in `design/choice-tape-shrinking.md`. In particular, fork and timeout configurations continue to use the `ValueTree` engine.

### DIFF
--- a/design/choice-tape-shrinking.md
+++ b/design/choice-tape-shrinking.md
@@ -416,11 +416,10 @@ strategy:
   `ShrinkEngine` config, runner integration. Tests: round-float minimal
   examples, filter no-stall, flat_map shrink quality, unmigrated-strategy
   shrinking via raw choices, budget respect.
-  **Status: DONE — see `design/choice-tape-status.md`.**
+  **Status: DONE.**
 - **Phase 2:** spans; continuation encoding for `vec`/collections;
   `delete_spans`; `prop_oneof` branch choice.
-  **Status: DONE except adaptive deletion batching — see
-  `design/choice-tape-status.md`.**
+  **Status: DONE, including adaptive deletion batching.**
 - **Phase 3:** migrate char/string/bits/sample; state-machine crate
   evaluation.
 - **Phase 4:** cross-value passes (duplicates, redistribute, reorder).
--- a/design/choice-tape-shrinking.md
+++ b/design/choice-tape-shrinking.md
@@ -418,7 +418,9 @@ strategy:
   shrinking via raw choices, budget respect.
   **Status: DONE — see `design/choice-tape-status.md`.**
 - **Phase 2:** spans; continuation encoding for `vec`/collections;
-  `delete_spans` with adaptive batching; `prop_oneof` branch choice.
+  `delete_spans`; `prop_oneof` branch choice.
+  **Status: DONE except adaptive deletion batching — see
+  `design/choice-tape-status.md`.**
 - **Phase 3:** migrate char/string/bits/sample; state-machine crate
   evaluation.
 - **Phase 4:** cross-value passes (duplicates, redistribute, reorder).
--- a/design/choice-tape-shrinking.md
+++ b/design/choice-tape-shrinking.md
@@ -1,0 +1,428 @@
+# Design: a Conjecture-style choice tape for proptest
+
+Status: draft for review. Nothing in this document is implemented yet.
+
+## 1. Motivation
+
+Proptest is "inspired by Hypothesis", but it did not adopt Hypothesis's core
+engine idea (Conjecture): serialize *all* generation decisions into a
+replayable sequence, and implement shrinking as *edits to that sequence*
+followed by re-running the generator. Instead, proptest threads shrink state
+through every strategy as a `ValueTree` with `simplify()`/`complicate()`
+([traits.rs](../proptest/src/strategy/traits.rs)), driven by a single greedy
+loop in [runner.rs](../proptest/src/test_runner/runner.rs) (`fn shrink`).
+
+Concrete costs of the current design, observed in the code:
+
+- **Floats shrink to noise.** `f64::BinarySearch`
+  ([num.rs](../proptest/src/num.rs), `float_bin_search!`) does numeric
+  bisection toward 0 and converges to an arbitrary boundary midpoint like
+  `0.7500000000000002`. Hypothesis shrinks through a lexicographic float
+  encoding and produces round values (`2.0`, `1.5`).
+- **`prop_filter` strands the search.**
+  [filter.rs](../proptest/src/strategy/filter.rs) rejects a simplification
+  whose value fails the predicate and backtracks; shrinking stalls at locally
+  minimal values (the code has to relax the simplify/complicate contract for
+  this, see `strict_complicate_after_simplify: false`).
+- **`prop_flat_map` shrinking is regeneration-with-retry.**
+  [flatten.rs](../proptest/src/strategy/flatten.rs) admits in its comments
+  that complicating after the outer value shrinks is implemented by
+  regenerating inner values up to a retry budget.
+- **No cross-value passes.** Each `ValueTree` shrinks in isolation. We cannot
+  lower two equal values together, redistribute a numeric pair, or reorder
+  elements.
+- **Fragile persistence.** Failures persist as RNG seeds
+  (`PersistedSeed(Seed)`), which regenerate a different value as soon as the
+  strategy code (or rand) changes. A persisted choice tape survives most code
+  changes because typed values are replayed, not re-derived from entropy.
+
+Decision (2026-07-07): build a choice-tape engine. Distribution changes to
+default strategies are acceptable.
+
+## 2. Goals and non-goals
+
+Goals, in priority order:
+
+1. Shrinking quality: round floats, no filter stalls, natural flat_map
+   shrinking, collection element deletion as a generic pass, later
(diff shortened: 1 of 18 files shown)


matthiasgoergens:
Here is a concrete demo of what this engine does to the oldest complaint class against proptest's shrinking: shrinking through `prop_flat_map`, the monadic bind.

**The problem, as reported by users.** #181 asks for "alternative approaches to `prop_flat_map` due to shrinking behavior": matrices generated as `(rows, cols).prop_flat_map(|(r, c)| vec(elem, r * c))` shrink their elements but rarely their dimensions. nalgebra's proptest support ships that caveat in its documentation to this day: "shrinking takes place by first shrinking the matrix elements before trying to shrink the dimensions of the matrix", so that "a large number of shrinking iterations are necessary to find a (nearly) minimal failing test case". The book's tutorial example for `prop_flat_map`, a vector plus a valid index into it, has the same shape.

**Why binds are hard for ValueTree.** A ValueTree shrinks the finished value, and a bind splits the value across two trees: the outer one (the dimensions) and an inner one built from its current value (the elements). Every outer shrink invalidates the inner tree's progress, so the search keeps restarting, and under the default iteration budget it often gives up with the outer value untouched. The workarounds discussed in #181 (custom ValueTree implementations, matrix-specific combinators) stay inside that paradigm. Hypothesis solved this class of problem at the engine level instead: shrink the recorded choices and replay generation, so the dependent structure regenerates consistently on every attempt. That is what this PR ports. One coordinated pass, taken from Hypothesis's `minimize_individual_nodes`, lowers a length choice while deleting one element span behind it, which is the edit neither side can make alone.

**Measured.** 100 fixed seeds per case, both engines from this branch, default config. "Fully minimal" counts runs that reached the ideal minimum stated with each case; "worst result" is that engine's furthest miss.

| case | engine | fully minimal | avg test calls | worst result |
|---|---|---|---|---|
| matrix, issue #181: `(1..=8, 1..=8)` flat_map `vec(0..1000, r*c)`, fail iff any elem >= 10 <br> ideal minimum: `(1, 1, [10])` | ValueTree | 95/100 | 405 | `(1, 8, [0, 0, 0, 0, 0, 6, 400, 703])` |
| | tape | 98/100 | 69 | `(2, 3, [0, 0, 0, 0, 0, 10])` |
| book's vec-and-index: `(1..=32)` flat_map `(vec(0..1000, len), 0..len)`, fail iff `v[i] >= 10` <br> ideal minimum: `([10], 0)` | ValueTree | 100/100 | 298 | |
| | tape | 100/100 | 51 | |
| length-prefixed sum: `(1..=64)` flat_map `vec(0..1000, len..=len)`, fail iff sum >= 100 <br> ideal minimum: `[100]` | ValueTree | 89/100 | 557 | `[0, 0, 0, 0, 0, 10, 393, 453]` |
| | tape | 100/100 | 109 | |
| chained binds: `a in 10..1000`, `b in 10..=a`, `c in 10..=b`, fail always <br> ideal minimum: `(10, 10, 10)` | ValueTree | 100/100 | 112 | |
| | tape | 100/100 | 2 | |

Two things stand out beyond the minimality counts. The tape gets there in a fraction of the test executions (4x to 56x fewer), which matters for expensive properties. And when either engine falls short of the global minimum, the failure modes differ in kind: ValueTree's non-minimal results carry leftover random values (`400`, `703`, `453`), while the tape's worst case is a correctly minimized value with surplus structure around it.

The `chained binds` row is the purest illustration: three dependent draws through two binds, and the tape's first shrink attempt (set every choice to its target) lands on `(10, 10, 10)` immediately, because replaying generation keeps the dependency `a >= b >= c` intact by construction.

<details>
<summary>Reproducible demo code (drop into <code>proptest/examples/flat-map-demo.rs</code> on this branch)</summary>

```rust
use std::cell::Cell;
use std::fmt::Debug;

use proptest::collection::vec;
use proptest::prelude::*;
use proptest::test_runner::{
    Config, RngAlgorithm, ShrinkEngine, TestError, TestRng, TestRunner,
};

fn run_case<S: Strategy>(
    engine: ShrinkEngine,
    strategy: &S,
    fail: &impl Fn(&S::Value) -> bool,
    seed_byte: u8,
) -> (Option<S::Value>, u64)
where
    S::Value: Debug,
{
    let calls = Cell::new(0u64);
    let seed = [seed_byte; 32];
    let mut runner = TestRunner::new_with_rng(
        Config {
            shrink_engine: engine,
            failure_persistence: None,
            ..Config::default()
        },
        TestRng::from_seed(RngAlgorithm::ChaCha, &seed),
    );
    let result = runner.run(strategy, |v| {
        calls.set(calls.get() + 1);
        if fail(&v) {
            Err(TestCaseError::fail("demo failure"))
        } else {
            Ok(())
        }
    });
    match result {
        Err(TestError::Fail(_, value)) => (Some(value), calls.get()),
        _ => (None, calls.get()),
    }
}

fn stats<S: Strategy>(
    name: &str,
    strategy: S,
    fail: impl Fn(&S::Value) -> bool,
    is_minimal: impl Fn(&S::Value) -> bool,
) where
    S::Value: Debug,
{
    println!("{} -- 100 seeds", name);
    for (label, engine) in [
        ("valuetree", ShrinkEngine::ValueTree),
        ("tape", ShrinkEngine::Tape),
    ] {
        let mut found = 0u32;
        let mut minimal = 0u32;
        let mut total_calls = 0u64;
        let mut worst: Option<String> = None;
        for seed_byte in 0..100u8 {
            let (value, calls) = run_case(engine, &strategy, &fail, seed_byte);
            total_calls += calls;
            if let Some(v) = value {
                found += 1;
                if is_minimal(&v) {
                    minimal += 1;
                } else {
                    let s = format!("{:?}", v);
                    if worst.as_ref().map_or(true, |w| s.len() > w.len()) {
                        worst = Some(s);
                    }
                }
            }
        }
        println!(
            "  {:<10} found {:>3}/100, fully minimal {:>3}/100, avg {:>4} test calls, worst: {}",
            label,
            found,
            minimal,
            total_calls / 100,
            worst.unwrap_or_else(|| "-".to_owned())
        );
    }
    println!();
}

fn main() {
    stats(
        "matrix (issue #181)",
        (1usize..=8, 1usize..=8).prop_flat_map(|(r, c)| {
            vec(0i32..1000, r * c).prop_map(move |data| (r, c, data))
        }),
        |(_, _, data)| data.iter().any(|&x| x >= 10),
        |(r, c, data)| *r == 1 && *c == 1 && *data == vec![10],
    );

    stats(
        "vec-and-index (book tutorial)",
        (1usize..=32)
            .prop_flat_map(|len| (vec(0i32..1000, len..=len), 0..len)),
        |(v, i)| v[*i] >= 10,
        |(v, i)| v.len() == 1 && v[0] == 10 && *i == 0,
    );

    stats(
        "length-prefixed sum",
        (1usize..=64).prop_flat_map(|len| vec(0i32..1000, len..=len)),
        |v| v.iter().sum::<i32>() >= 100,
        |v| *v == vec![100],
    );

    stats(
        "chained binds",
        (10i32..1000)
            .prop_flat_map(|a| (Just(a), 10..=a))
            .prop_flat_map(|(a, b)| (Just(a), Just(b), 10..=b)),
        |_| true,
        |t| *t == (10, 10, 10),
    );
}
```

</details>

The coordinated lower-and-delete pass and issue #181's matrix case in the `shrink-quality` example are the two most recent commits on the branch; there is a regression test asserting that the length-prefixed case shrinks to exactly `[100]`.

PR #662, opened 2026-07-23
Upgrade to `bit-set v0.11` and `bit-vec v0.10`

cuviper:


matthew-russo (approved):

Manishearth:
Can we get a crates.io release with this?

